PR #15, opened 2021-12-14
feat: implement Auction consumer

ThomasRalee:
## This PR:
- creates exchange auction consumer to be used with auction-v2.


packages/exchange-consumer/src/transformers/DMMTransformer.ts
```diff
@@ -192,6 +192,7 @@ export class DMMTransformer {
       vcs: vcs
         ? DMMTransformer.grpcVCSResultRecordToVCSResultRecords(vcs)
         : undefined,
+      createdAt: '', // epochResultRecord.getCreatedAt(),
```
ThomasRalee, line 195: Did some hack here to ensure the build pass, will update once the `@injectivelabs/exchange-api` library is updated.

PR #17, opened 2021-12-31
feat: xbx

XBurnX:


bangjelkoski:
Published, you can now use `@injectivelabs/token-metadata@0.0.35`.

PR #42, opened 2022-06-14
chore: add volume property to FeeDiscountTierInfo interface

afterburn:
As per the response for [this endpoint](https://devnet-1.lcd.injective.dev/injective/exchange/v1beta1/fee_discount_account_info/inj1hkhdaj2a2clmq5jq6mspsggqs32vynpk228q3r) we need a volume property.

### DIFF
--- a/deprecated/chain-consumer/src/transformers/ExchangeTransformer.ts
+++ b/deprecated/chain-consumer/src/transformers/ExchangeTransformer.ts
@@ -76,7 +76,9 @@ export class ExchangeTransformer {
       makerDiscountRate: info.getMakerDiscountRate(),
       takerDiscountRate: info.getTakerDiscountRate(),
       stakedAmount: info.getStakedAmount(),
-      volume: info.getVolume(),
+      // @ts-ignore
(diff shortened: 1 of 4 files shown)


ThomasRalee (dismissed):

shane-moore (dismissed):

bangjelkoski:
Adding just the type to the interface doesn't do anything, you need to also add that field in the transformers (https://github.com/InjectiveLabs/injective-ts/blob/9bb7d860af7bad5ace82de37dc3b1bf56501d693/packages/sdk-ts/src/client/chain/transformers/ChainGrpcExchangeTransformer.ts#L125-L138) 

cc @ThomasRalee @shane-moore

bangjelkoski:
We should not merge this now and republish as it will break the current flow - this update is only for after the chain upgrade is completed.

bangjelkoski (approved):

bangjelkoski:
```
Successfully published:
 - @injectivelabs/sdk-ts@1.0.29
 - @injectivelabs/sdk-ui-ts@1.0.53
```
 
 @afterburn

PR #48, opened 2022-07-05
fix: add moonbeam logo to token-metadata

afterburn:


afterburn:
@ThomasRalee is looking into this some more, so for now I've turned this into a draft PR.

PR #49, opened 2022-07-06
feat: add support for CW20 balance

ThomasRalee:
## Changes
- updated sdk-ts/sdk-ui-ts/token-metadata to support the `Token` type's new `TokenType` property, which can be used in any product to differentiate between `ERC20/CW20/Native/IBC/InsuranceFund` tokens
- `isCanonical` property added to `IBCToken` type

## Todo
- after new drawings are finished, icons for `InsuranceFund`, `non-canonical IBC`, and `CW20` tokens will be added to `token-metadata`, and logic to select the appropriate icon based on `tokenType` will be updated as well
- update hub, explorer, and dex/helix to use new packages. this is pretty much done in my existing `hub` and `explorer` PR's. I only saw one small change that will be required for `dex/helix`, so I can knock that out after these are published


packages/sdk-ui-ts/src/utils/helpers.ts
```diff
@@ -15,3 +16,15 @@ export const getTokenLogoWithVendorPathPrefix = (image: string) => {
 
   return path.join('/', 'vendor', '@injectivelabs', 'token-metadata', image)
 }
+
+export const getTokenTypeBaseOnAddress = (address: string): TokenType => {
+  if (address.startsWith('inj')) {
+    return TokenType.CW20
```
bangjelkoski, line 22: Why is the native denom type CW20? We can maybe have another TokenType - bankDenom or NativeToken or something like that

bangjelkoski (changes requested):

packages/sdk-ui-ts/src/utils/helpers.ts
shane-moore: I wiped out this method since it's no longer useful. it is being removed from hub and explorer and replaced by just checking the `TokenType` property of a denom

packages/sdk-ui-ts/src/types/token.ts
```diff
@@ -32,6 +32,9 @@ export interface CoinPriceFromInjectiveService {
 export interface TokenWithBalance extends Token {
   balance: string
   allowance: string
+  baseDenom?: string
```
ThomasRalee, line 35: Let's use the implementation for `BankBalanceWithToken` instead of re declaring: 
```
{
    baseDenom?: string
    channelId?: string
    isCanonical?: boolean
}
```

shane-moore: @ThomasRalee `BankBalanceWithToken` is a different type than what we're accomplishing with `TokenWithBalance`. `TokenWithBalance` contains an `allowance` property, which is not present on `BankBalanceWithToken`. I agree that our `token` typings could use some work, and I would like to take a crack at it in the next PR

packages/sdk-ui-ts/src/token/TokenService.ts
```diff
@@ -299,7 +301,21 @@ export class TokenService {
       .replaceAll('/', '-')
       .replaceAll(' ', '-')
       .toLowerCase()
+
     const [baseTokenSymbol] = market.ticker.replace(quoteToken.symbol, '')
+
+    let tokenType
```
bangjelkoski, line 307: Extract this to a function, `getTokenTypeFromSymbol` or something like that. I also don't like that we have to classify the `baseToken` for derivative markets; they don't have such a concept since there is no `baseDenom` for derivative markets; we've added this so that it's easier for us to render some stuff on the UI level. 

Please rethink the approach here

bangjelkoski (changes requested):

packages/sdk-ui-ts/src/token/TokenService.ts
shane-moore: @bangjelkoski , yeah, I thought about how there is not a real `baseToken` in a derivatives market as well. and you're right that we're just doing it to support the concept of a `baseAmount` in a perp market and that we can have a cleaner implementation.

for this function `getBinaryOptionsMarketWithToken`, it is not currently used in the dex, since we're not currently supporting any binaryOptions markets there.  is there a UI I can use to test a new implementation of the method? perhaps make a feature branch off of Front Runner's repo and hook up to one of their markets. don't see a link to their github repo on their site though so perhaps they're closed source atm

PR #53, opened 2022-07-28
chore: create getTokenLogoFromTokenType helper method and add error handling for fetchCW20Balances

shane-moore:
This PR addresses the concerns noted [here](https://github.com/InjectiveLabs/injective-explorer/pull/99#discussion_r931894220) and [here](https://github.com/InjectiveLabs/injective-hub/pull/661#discussion_r932157698).

### DIFF
--- a/packages/sdk-ts/src/client/exchange/rest/ExchangeRestExplorerApi.ts
+++ b/packages/sdk-ts/src/client/exchange/rest/ExchangeRestExplorerApi.ts
@@ -1,4 +1,4 @@
-import { TokenType } from '@injectivelabs/token-metadata'
+import BaseRestConsumer from '../../BaseRestConsumer'
 import {
   BlockFromExplorerApiResponse,
   ContractExplorerApiResponse,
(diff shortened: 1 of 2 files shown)


packages/sdk-ui-ts/src/utils/helpers.ts
```diff
@@ -15,3 +16,29 @@ export const getTokenLogoWithVendorPathPrefix = (image: string) => {
 
   return path.join('/', 'vendor', '@injectivelabs', 'token-metadata', image)
 }
+
+export const getTokenLogoFromTokenType = (token: Token) => {
+  const { tokenType, logo } = token
+
+  const { isCanonical } = token as IbcToken
```
bangjelkoski, line 23: No need for a blank row above

bangjelkoski (changes requested):

packages/sdk-ts/src/client/exchange/rest/ExchangeRestExplorerApi.ts
```diff
@@ -355,6 +355,26 @@ export class ExchangeRestExplorerApi extends BaseRestConsumer {
         ExchangeRestExplorerTransformer.CW20BalanceToExplorerCW20Balance,
       )
     } catch (error: any) {
+      if (error.message.includes(404) || error.message.includes(500)) {
```
bangjelkoski, line 358: We should have a `fetchCW20BalancesNoThrow` version of the function where the error is handled and use that where we need to silently ignore these errors. Also, are we sure we should return a blank contract address and not `undefined`?

bangjelkoski (changes requested):

packages/sdk-ui-ts/src/utils/helpers.ts
```diff
@@ -15,3 +16,28 @@ export const getTokenLogoWithVendorPathPrefix = (image: string) => {
 
   return path.join('/', 'vendor', '@injectivelabs', 'token-metadata', image)
 }
+
+export const getTokenLogoFromTokenType = (token: Token) => {
```
ThomasRalee, line 20: Let's move the non-hardcoded logo to the UI, which will give us some room for flexibility. Example: CW20 logo could be a byte64 string, and we might want to do some kind of customization to render them the the user.

```
case tokenType === TokenType.Cw20 && logo:
      return logo
case isIbcOrEthTokenWithVendorLogo:
      return getTokenLogoWithVendorPathPrefix(logo)
```

packages/sdk-ts/src/client/exchange/rest/ExchangeRestExplorerApi.ts
shane-moore: updated to add a fetchCW20BalancesNoThrow method. regarding `contractAddress`, I discussed with @ThomasRalee  earlier, and `contractAddress` should always be string - since we wont have cw20 balance with undefined contractAddress

packages/sdk-ui-ts/src/utils/helpers.ts
shane-moore: @ThomasRalee I removed the non-hardcoded logo back to the UI level

ThomasRalee (approved):

PR #56, opened 2022-08-11
chore: funding rates pagination support

shane-moore:
##Changes
- updated indexer-api to 1.0.4
- added pagination support to `fetchFundingRates`

### DIFF
--- a/packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcDerivativesApi.ts
+++ b/packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcDerivativesApi.ts
@@ -469,8 +469,9 @@ export class IndexerGrpcDerivativesApi extends BaseConsumer {
         DerivativeSubaccountTradesListResponse,
         typeof InjectiveDerivativeExchangeRPC.SubaccountTradesList
       >(request, InjectiveDerivativeExchangeRPC.SubaccountTradesList)
-
-      return IndexerGrpcDerivativeTransformer.tradesResponseToTrades(response)
+      return IndexerGrpcDerivativeTransformer.tradesResponseToTrades(
+        response as DerivativeTradesResponse,
+      )
     } catch (e: any) {
       throw new Error(e.message)
     }
--- a/packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcDerivativesApi.ts
+++ b/packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcDerivativesApi.ts
@@ -469,9 +469,7 @@ export class IndexerGrpcDerivativesApi extends BaseConsumer {
         DerivativeSubaccountTradesListResponse,
         typeof InjectiveDerivativeExchangeRPC.SubaccountTradesList
       >(request, InjectiveDerivativeExchangeRPC.SubaccountTradesList)
-      return IndexerGrpcDerivativeTransformer.tradesResponseToTrades(
-        response as DerivativeTradesResponse,
-      )
+      return IndexerGrpcDerivativeTransformer.tradesResponseToTrades(response)
     } catch (e: any) {
       throw new Error(e.message)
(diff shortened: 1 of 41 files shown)


packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcDerivativesApi.ts
```diff
@@ -469,8 +469,9 @@ export class IndexerGrpcDerivativesApi extends BaseConsumer {
         DerivativeSubaccountTradesListResponse,
         typeof InjectiveDerivativeExchangeRPC.SubaccountTradesList
       >(request, InjectiveDerivativeExchangeRPC.SubaccountTradesList)
-
-      return IndexerGrpcDerivativeTransformer.tradesResponseToTrades(response)
+      return IndexerGrpcDerivativeTransformer.tradesResponseToTrades(
+        response as DerivativeTradesResponse,
```
shane-moore, line 473: @bangjelkoski I was getting typing error for the `response` parameter stating that it needs `response` as type `DerivativeTradesResponse`. curious what you think here. i'm also curious if this has anything to do with the dex failing earlier on `DerivativeTradesRequest and SpotTradesRequest`

packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcSpotApi.ts
```diff
@@ -288,8 +288,9 @@ export class IndexerGrpcSpotApi extends BaseConsumer {
         SpotSubaccountTradesListResponse,
         typeof InjectiveSpotExchangeRPC.SubaccountTradesList
       >(request, InjectiveSpotExchangeRPC.SubaccountTradesList)
-
-      return IndexerGrpcSpotTransformer.tradesResponseToTrades(response)
+      return IndexerGrpcSpotTransformer.tradesResponseToTrades(
+        response as SpotTradesResponse,
```
shane-moore, line 292: @bangjelkoski ^^ same comment as above

packages/sdk-ts/src/client/indexer/types/derivatives.ts
```diff
@@ -191,6 +192,11 @@ export interface FundingRate {
   timestamp: number
 }
 
+export interface FundingRateWithPagination {
```
bangjelkoski, line 195: Not the pattern we do, you don't need to specify the exact interface it returns, or if you want you can just do it as 
```ts
export const transformer = (response: Response): {
  fundingRates: FundingRate[]
  pagination: ExchangePagination
} => { /* */ }
```

packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcSpotApi.ts
bangjelkoski: We should try and resolve this without typecasting (same for above. Maybe the import in the `IndexerGrpcSpotTransformer` is from the `exchange-api` instead?

bangjelkoski (changes requested):

packages/sdk-ts/src/client/indexer/types/derivatives.ts
shane-moore: @bangjelkoski got it, I destroyed this interface

packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcSpotApi.ts
shane-moore: @bangjelkoski this was resolved by merging in `master` branch, which updated the method call to `IndexerGrpcSpotTransformer.subaccountTradesListResponseToTradesList(response)`, so we're all good here. also was resolved for `IndexerGrpcDerivativeTransformer`

bangjelkoski (approved):

PR #57, opened 2022-08-11
WIP:  Auths module - add MsgExec and new param MsgGrant

Grinion:
Hello,

I added a new message type (MsgExec) to authz module and add a new param expiryInSeconds for MsgGrant

Also, I think we need to delete param expiryInYears. What do you think about it?

### DIFF
--- a/packages/sdk-ts/src/core/authz/msgs/MsgExec.ts
+++ b/packages/sdk-ts/src/core/authz/msgs/MsgExec.ts
@@ -1,13 +1,13 @@
 import { MsgExec as BaseMsgExec } from '@injectivelabs/chain-api/cosmos/authz/v1beta1/tx_pb'
-import { Message } from 'google-protobuf'
 import { Any } from 'google-protobuf/google/protobuf/any_pb'
 
 import { MsgBase } from '../../MsgBase'
+import { Msgs } from '../../msgs'
 
 export declare namespace MsgExec {
   export interface Params {
     grantee: string
-    msgs: MsgBase<unknown, any, Message, unknown, unknown> | MsgBase<unknown, any, Message, unknown, unknown>[]
+    msgs: Msgs | Msgs[]
   }
 
   export interface DirectSign {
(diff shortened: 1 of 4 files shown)


packages/sdk-ts/src/core/authz/msgs/MsgExec.ts
```diff
@@ -0,0 +1,91 @@
+import { MsgExec as BaseMsgExec } from '@injectivelabs/chain-api/cosmos/authz/v1beta1/tx_pb'
+import { Message } from 'google-protobuf'
+import { Any } from 'google-protobuf/google/protobuf/any_pb'
+
+import { MsgBase } from '../../MsgBase'
+
+export declare namespace MsgExec {
+  export interface Params {
+    grantee: string
+    msgs: MsgBase<unknown, any, Message, unknown, unknown> | MsgBase<unknown, any, Message, unknown, unknown>[]
```
bangjelkoski, line 10: These should be a separate abstraction class, similar to what we have to ExecArgs for cosmwasm contracts, you can check a reference codebase [here](https://github.com/InjectiveLabs/injective-ts/tree/master/packages/sdk-ts/src/core/wasm/exec-args)

packages/sdk-ts/src/core/authz/msgs/MsgExec.ts
```diff
@@ -0,0 +1,91 @@
...
+    return new MsgExec(params)
+  }
+
+  toProto(): MsgExec.Proto {
+    const { params } = this
+
+    const message = new BaseMsgExec()
+    message.setGrantee(params.grantee)
+
+    if (Array.isArray(params.msgs)) {
```
bangjelkoski, line 46: Here, instead of having these `if` conditions we can do 

```ts
const msgs = Array.isArray(params.msgs) ? msgs : [msgs]

params.msgs.forEach(msg => {
    const msgValue = new Any()
    msgValue.setTypeUrl(msg.toData()['@type'])
    msgValue.setValue(msg.toProto().serializeBinary())

    message.addMsgs(msgValue)
  });
  ```

**Review by bangjelkoski — CHANGES_REQUESTED** [review]

File: packages/sdk-ts/src/core/authz/msgs/MsgExec.ts (hunk 1 above)
Comment (Grinion) [reply packages/sdk-ts/src/core/authz/msgs/MsgExec.ts:10]: Sorry, what do you mean? Add for every message class ?

Comment (bangjelkoski) [reply packages/sdk-ts/src/core/authz/msgs/MsgExec.ts:10]: Yes, you can start by adding the one that you need and then we can extend them further depending on the usecase

Comment (Grinion) [reply packages/sdk-ts/src/core/authz/msgs/MsgExec.ts:10]: I need all messages and the further remember that need to add I think it's no good way  :( 

Maybe ? 
import { Msgs , ExchangeMsgs,} from '../../msgs'

msgs: Msgs | ExchangeMsgs

Comment (bangjelkoski) [reply packages/sdk-ts/src/core/authz/msgs/MsgExec.ts:10]: Oh, sorry, yeah I misunderstood what `msgs` mean here - let's use `msgs: Msgs | Msgs[]` here.

**Review by bangjelkoski — APPROVED** [review]

**Comment by bangjelkoski** [comment]
I'll comment her ethe updated packages versions by the end of the day

**Comment by bangjelkoski** [comment]
```
Successfully published:
 - @injectivelabs/sdk-ts@1.0.82
 - @injectivelabs/sdk-ui-ts@1.0.119
 ```
 
 cc @Grinion

PR #78, opened 2022-09-15
fix: added executionSide prop to trade objects

afterburn:
This PR makes the executionSide property visible on trade objects so we can use it to filter for either `maker` or `taker` side in our products.


bangjelkoski (approved):

bangjelkoski:
Updated, use `ncu | grep @injective` and update to the latest packages, most of them got updated because there was a change on the `@injectivelabs/ts-types` package. @afterburn

PR #79, opened 2022-09-20
feat/exceptions

bangjelkoski:


ThomasRalee (approved):

albertchon:
@bangjelkoski perhaps these errors/error strings should be extracted out to a separate file? It would make updating the copy much easier in the future. Error copy is important

bangjelkoski:
I need to add the rest of the CosmosSDK native ones + from our insurance module and we are good to go. Then, the product guys can decide on the messages we will show to the user.

PR #85, opened 2022-09-27
fix: RN GRPC issue

billyjacoby:


packages/sdk-ts/src/utils/helpers.ts
```diff
@@ -20,6 +20,10 @@ export const isNode = () => {
 }
 
 export const isBrowser = () => {
+  if (isReactNative()) {
+    return false
+  }
+
```
billyjacoby, line 26: We have to check for this before checking if window is defined since there are cases in RN where window is polyfilled.

bangjelkoski:
```
Successfully published:
 - @injectivelabs/sdk-ts@1.0.163
 - @injectivelabs/sdk-ui-ts@1.0.229
 - @injectivelabs/wallet-ts@1.0.129
 ```

PR #98, opened 2022-12-05
feat: utility methods for explorer

shane-moore:
## Changes
Utility method `getBlockExplorerPathFromNetworkType` can now be used across products to grab the appropriate block explorer url for a given address across Injective, Ethereum, and Cosmos block explorers. It can also easily be extended to other chains such as Solana.


bangjelkoski:
What is the use case for this to implement right now, and where are we going to use it?

PR #100, opened 2022-12-07
fix: fixes walletconnect signed tx type

billyjacoby:
Updates the type of request sent to Wallet Connect to allow for Metamask and many other wallets to properly sign transactions


bangjelkoski:
New package versions published!

PR #103, opened 2022-12-09
feat: bridge support for sommelier

shane-moore:
## Changes
- add support for Sommelier bridge transfers through IBC
- token metadata for ETHBTC Trend, Steady ETH, Steady BTC, and Sommelier

Note: We are missing the gas price for Sommelier. @achilleas-kal is asking their team for this.


packages/token-metadata/src/ibc/tokens.ts
```diff
@@ -63,4 +63,8 @@ export default {
     path: 'transfer/channel-13',
     baseDenom: 'ASTRO',
   },
+  '34346a60a95eb030d62d6f5bdd4b745be18e8a693372a8a347d5d53dbbb1328b': {
```
bangjelkoski, line 66: all uppercase please

bangjelkoski (changes requested):

packages/token-metadata/src/ibc/tokens.ts
shane-moore: @bangjelkoski updated to uppercase 🤠

bangjelkoski (approved):

PR #122, opened 2023-01-25
feat: map chain error exception by module

ThomasRalee:
## Description:
Currently, error codes are duplicated between modules, resulting in UI showing the wrong error toast. This PR maps chain error exceptions by the module type, ensuring that the toast content is accurate to the chain error.

### DIFF
--- a/packages/exceptions/src/exceptions/TransactionException.ts
+++ b/packages/exceptions/src/exceptions/TransactionException.ts
@@ -12,10 +12,17 @@ export class TransactionException extends ConcreteException {
   public parse(): void {
     const { message } = this
 
-    const { message: parsedMessage, code } =
-      mapFailedTransactionMessage(message)
+    const {
+      message: parsedMessage,
+      code,
+      module,
+    } = mapFailedTransactionMessage(message)
 
     this.setMessage(parsedMessage)
     this.setContextCode(code)
+
+    if (module) {
+      this.setContextModule(module)
+    }
   }
 }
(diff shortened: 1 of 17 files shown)


packages/exceptions/src/utils/maps.ts
```diff
@@ -1,88 +1,57 @@
...
 
 export const mapFailedTransactionMessage = (
   message: string,
-): { message: string; code: ErrorContextCode } => {
+): { message: string; code: ErrorContextCode; module?: string } => {
   const ABCICode = getABCICode(message)
+  const module = getContextModule(message)
+
+  const defaultResponse = {
+    message: 'The request is not known',
```
bangjelkoski, line 35: Let's change this to something better, maybe `Transaction execution has failed for unknown reason` or something similar

packages/exceptions/src/exceptions/TransactionException.ts
```diff
@@ -12,10 +12,17 @@ export class TransactionException extends ConcreteException {
...
-    const { message: parsedMessage, code } =
-      mapFailedTransactionMessage(message)
+    const {
+      message: parsedMessage,
+      code,
+      module,
+    } = mapFailedTransactionMessage(message)
 
     this.setMessage(parsedMessage)
     this.setContextCode(code)
```
bangjelkoski, line 22: In some cases we pass the code within the constructor, like [here](https://github.com/InjectiveLabs/injective-ts/blob/3947185308da9ed553ad6686b281d87fcbeec889/packages/sdk-ts/src/core/modules/tx/api/TxRestApi.ts#L58-L61) so we should just use that instead of using the parsed one

bangjelkoski (changes requested):

PR #128, opened 2023-02-13
chore: add explorer link based on tsx for eth wh transfer

shane-moore:
This PR makes it possible to get explorer link URL's attached to wormhole ethereum deposits.

### DIFF
--- a/packages/sdk-ui-ts/src/client/transformers/UiBridgeTransformer.ts
+++ b/packages/sdk-ui-ts/src/client/transformers/UiBridgeTransformer.ts
@@ -25,6 +25,7 @@ import { UserDeposit } from '@injectivelabs/sdk-ts/dist/client'
 import {
   getExplorerUrl,
   getCosmosExplorerUrl,
+  getSolanaExplorerUrl,
   getEthereumExplorerUrl,
   getTerraExplorerUrl,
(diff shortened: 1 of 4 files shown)


packages/sdk-ui-ts/src/client/transformers/UiBridgeTransformer.ts
```diff
@@ -346,17 +346,31 @@ export const convertMoonbeamToUiBridgeTransaction = async (
...
-  transaction: WormholeTxResponse,
-): Promise<UiBridgeTransaction> => {
+export const convertWormholeToUiBridgeTransaction = async ({
+  transaction,
+  network,
+}: {
+  transaction: WormholeTxResponse
+  network: Network
+}): Promise<UiBridgeTransaction> => {
+  const isEthereumWhTransfer =
```
bangjelkoski, line 356: Why don't we check also by the tx.source/tx.destination?

shane-moore: ahh good call! because then we can check explicitly that it's an ethereum WH transfer, as opposed to a peggy eth transfer for example added in latest commit

bangjelkoski (approved):

PR #130, opened 2023-02-14
fix: factory token decimals

shane-moore:
## Changes
This PR attempts to resolve the scenario mentioned in https://www.notion.so/injective/91856a6f88874b03b9098f6980e0414f?v=6d230a442a594f05af3667d8f32a5759&p=a51e702fce70471ba566ba90854f2aba&pm=s.


bangjelkoski (approved):

bangjelkoski:
This is fine, but like I said there is no need for usage of this function in production as we can just directly use `cw20.decimals`.

PR #132, opened 2023-02-17
chore: helper method for Solana getTxResponse

shane-moore:
## Changes
- the new `getTxResponse` for the `SolanaWormholeClient` will fetch a `TransactionResponse` based on a passed in `txHash.
- `RequestingVAA`  was added to the `BridgeTransactionState` in order to mark a transaction on the hub's bridge as being in a state of about to request a VAA. Therefore, if the user refreshes the page, they can see their order in the order details table, and the flow will go back to the point of trying to request a VAA from the WH guardians so that their transfer can be redeemed on the appropriate network end.

### DIFF
--- a/packages/bridge-ts/src/wormhole/clients/SolanaWormholeClient.ts
+++ b/packages/bridge-ts/src/wormhole/clients/SolanaWormholeClient.ts
@@ -196,6 +196,24 @@ export class SolanaWormholeClient extends WormholeClient {
     }
   }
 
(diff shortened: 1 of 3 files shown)


bangjelkoski (approved):

bangjelkoski:
Successfully published:
 - @injectivelabs/bridge-ts@1.0.275
 - @injectivelabs/sdk-ui-ts@1.0.571

PR #133, opened 2023-02-17
Chore/tens multiplier

shane-moore:
## Changes
The `formatNumberToAllowableTensMultiplier` helper function has been modified to round to a tens value based upon a passed in rounding mode. 

For example:
value: `69`
tensMultiplier: `10`
roundingMode: `round_down`

The result will now be `60` in lieu of previously `70`.  

The reason being is that if a user enters in an amount of 69, we don't want to round it up, because they will end up paying more for an extra quantity.

### DIFF
--- a/packages/bridge-ts/CHANGELOG.md
+++ b/packages/bridge-ts/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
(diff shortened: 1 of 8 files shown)


bangjelkoski (approved):

bangjelkoski:
```
Successfully published:
 - @injectivelabs/bridge-ts@1.0.276
 - @injectivelabs/sdk-ts@1.0.447
 - @injectivelabs/sdk-ui-ts@1.0.572
 - @injectivelabs/wallet-ts@1.0.446
 ```

PR #137, opened 2023-02-21
refactor: token metadata typing

ThomasRalee:
## This PR:
- refactors toke meta typing

### DIFF
--- a/packages/token-metadata/src/types.ts
+++ b/packages/token-metadata/src/types.ts
@@ -23,6 +23,7 @@ export interface IbcTokenMeta {
   symbol?: string
   isNative: boolean
   baseDenom: string
+  tokenType: TokenType.Ibc
 }
 
 export interface SplTokenMeta {
@@ -43,6 +44,8 @@ export interface Cw20TokenMeta {
   address: string
   decimals: number
   tokenType: TokenType.Cw20
+  symbol?: string
+  source?: Cw20TokenSource
 }
 
 export interface Cw20TokenMetaWithSource extends Cw20TokenMeta {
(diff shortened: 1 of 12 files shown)


packages/token-metadata/src/types.ts
```diff
@@ -59,9 +62,9 @@ export interface TokenMeta {
   coinGeckoId: string
 
   ibc?: IbcTokenMeta
+  ibcs?: IbcTokenMeta[]
   spl?: SplTokenMeta
-  cw20?: Cw20TokenMeta
-  cw20s?: Cw20TokenMetaWithSource[] // When there are multiple variations of the same CW20 token
+  cw20s?: Cw20TokenMeta[]
```
bangjelkoski, line 67: Why do we have`ibc?:` but we dont have `cw20?:`?

bangjelkoski (changes requested):
Can we check if there are some other places on the `injective-ts` wether we are accessing `token.cw20` or `token.ibc`?

packages/token-metadata/src/types.ts
ThomasRalee: I've made it so that cw20 meta is merged to the main token meta body, we can directly use it as per any other tokens in our implementation.

So instead of `token.cw20.decimals`, we can use `token.decimals` directly. 

For cw20 source, we can do this:
```
(token as Cw20Token).soruce
```
For ibc, because of the list of custom fields, it will be cleaner to introduce a separate type as the format we have now:
`token.ibc.baseDenom`

bangjelkoski: I think we should keep it consistent, either have both ibc and cw20 (which I'd prefer) or both merged

ThomasRalee: sure, will revert

bangjelkoski (approved):

PR #155, opened 2023-03-21
Fix eip712 for date type and feePayer

Grinion:
Hello,

These changes have been made to fix the MsgGrant message
https://github.com/InjectiveLabs/injective-ts/blob/master/packages/sdk-ts/src/core/modules/authz/msgs/MsgGrant.ts

I'm not sure about this change. Maybe this condition has been added for different messages. Do you know?
```
-   if (property === '@type' || property === 'type') {
+    if (property === '@type') {
```

Also, I have fixed feepayer for transaction of type eip 712

### DIFF
--- a/packages/sdk-ts/src/core/modules/authz/msgs/MsgGrant.spec.ts
+++ b/packages/sdk-ts/src/core/modules/authz/msgs/MsgGrant.spec.ts
@@ -1,0 +1,121 @@
+import { MsgGrant as BaseMsgGrant } from '@injectivelabs/chain-api/cosmos/authz/v1beta1/tx_pb'
+import MsgGrant from './MsgGrant'
+import { mockFactory } from '@injectivelabs/test-utils'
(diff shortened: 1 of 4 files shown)


bangjelkoski:
> > I'd love if you guys add unit tests for these messages if you are able.
> 
> Done

Are we good on their other AuthZ messages as well? It's been a while since we've used them so (I know I'm picky here) it'd be of great help if you can add tests of them as well. I'll merge this and redeploy anyway but just if you are up for it :)

bangjelkoski:
```
Successfully published:
 - @injectivelabs/bridge-ts@1.0.327
 - @injectivelabs/sdk-ts@1.0.498
 - @injectivelabs/sdk-ui-ts@1.0.632
 - @injectivelabs/wallet-ts@1.0.504
 ```
 
 Once more again - If you can write tests for the other Authz messages it'll be awesome. Cheers

PR #156, opened 2023-03-21
Add signer pubkey type

Grinion:
Each chain can have several different key types.  For some case,  I use '/cosmos.crypto.secp256k1.PubKey' for Injective chain
 

I'm also considering making a new pubKeyType field - pass the Proto Any object directly to the pubKey field. What do you think about it?


bangjelkoski:
The current implementation is closed for modification but open for extension. You could wrap the current implementation in a new one that could fit your needs.

PR #169, opened 2023-04-08
adding SCV-Security logo

unl1k3ly:


packages/sdk-ui-ts/src/utils/mappings.ts
```diff
@@ -314,5 +314,7 @@ export const validatorAddressToPathMap = {
     injvalcons1n5yluh6k9zqv74jw35r5s5zy808xpzfqp2wm7g:
     'injvaloper1n5yluh6k9zqv74jw35r5s5zy808xpzfq4ea8jf.webp',
     injvaloper1n5yluh6k9zqv74jw35r5s5zy808xpzfq4ea8jf:
-    'injvaloper1n5yluh6k9zqv74jw35r5s5zy808xpzfq4ea8jf.webp'
```
bangjelkoski, line 317: Why is this removed?

bangjelkoski (changes requested):

packages/sdk-ui-ts/src/utils/mappings.ts
bangjelkoski: You are also missing the `injvalcons` mapping

PR #171, opened 2023-04-14
chore: validator images script

shane-moore:
## Changes
`validators-logo` files were moved out of the `src` directory in `sdk-ui-ts`.

A script was created that will query `keybase` for the url path to the validator images. Then, a separate query is made to fetch the images. The script will store the images in a new `validators-logo/images` folder which also houses the existing manually inputted validator images.  The reason for this is that there should be a fall back image for a validator in the scenarios where the validator does not have an image stored via `keybase`.  This also leads to redundant images, so `sdk-ui-ts` will now ship with almost twice as many image files, mostly `jpeg`.  If we want to reduce the bundle size, we can manually delete all the redundant manually inputted images and rely on `keybase` for them. We will just lose the fallback option.  

A json mappings file is also output from the script that will map the validator's address to the location of the `keybase` image file.  If the validator's address is found in the json list, it's validator image will be used, else the manually created image will be used.  You can check out how the fallback works via the modifications to `getValidatorLogoWithVendorPathPrefix`.  

There are multiple cli commands that can be used for different use cases:
- default `yarn validators-logo` will only update the validators logo json map if a validator now has an `identifier` to send to keybase but didn't previously. Then, it will fetch the image.

- `yarn validators-logo:update-all` can be used to recreate the keybase json map and refetch all of the images. This is useful to handle scenarios where validators have updated their already existing validator image with a new one.

-`yarn validators-logo:update-individual <injvaloperAddress>` can be used to update a single validator on the keybase json map and refetch it's corresponding image

The consensus nodes address (`injvalcons..`) was not added to the keybase map because none of the products pass this as a param to `getValidatorLogoWithVendorPathPrefix` but can be added if useful for some reason I'm not considering

## References
- [notion](https://www.notion.so/injective/Frontend-Team-Dashboard-cfa9a0cacd1c40afb5ab399d68bae83c?p=4fadde3711f3430bb5fb875e8ae2c109&pm=s)


packages/sdk-ui-ts/validators-logo/index.ts
```diff
@@ -0,0 +1,134 @@
...
+
+const KEYBASE_SUFFIX = '_/api/1.0/user/lookup.json?fields=pictures&key_suffix='
+
+/* create validator to identity map */
+const fetchValidators = async () => {
+  const { validators } = await chainGrpcStakingApi.fetchValidators()
+
+  return validators.reduce((acc: ValidatorMap, validator: Validator) => {
+    acc[validator.operatorAddress] = validator.description.identity
+    return acc
```
ThomasRalee, line 24: we can use spread syntax here
```
=> { ...acc, [validator.operatorAddress]: validator.description.identity }
```

packages/sdk-ui-ts/validators-logo/index.ts
```diff
@@ -0,0 +1,256 @@
...
+const KEYBASE_SUFFIX = 'user/lookup.json?fields=pictures&key_suffix='
+
+const existingValidatorToAddressMapFromKeybase =
+  { ...JSON.parse(JSON.stringify(validatorToAddressMapFromKeybase)) } || {}
+
+/* cli flags */
+const [, , flag] = process.argv
+const cliValidatorAddress = flag && flag.includes('inj') ? flag : undefined
+const shouldUpdateAllImages = flag && flag.includes('update:all')
+const removeImages = flag && flag.includes('remove:images')
```
ThomasRalee, line 27: Lets remove this flag and the relevant codes. 2 reasons:
1. this is a flag we will use very rarely
2. we can always manually remove the images & mapping

packages/sdk-ui-ts/validators-logo/index.ts
```diff
@@ -0,0 +1,256 @@
...
+
+const KEYBASE_SUFFIX = 'user/lookup.json?fields=pictures&key_suffix='
+
+const existingValidatorToAddressMapFromKeybase =
+  { ...JSON.parse(JSON.stringify(validatorToAddressMapFromKeybase)) } || {}
+
+/* cli flags */
+const [, , flag] = process.argv
+const cliValidatorAddress = flag && flag.includes('inj') ? flag : undefined
+const shouldUpdateAllImages = flag && flag.includes('update:all')
```
ThomasRalee, line 26: Lets rename the flags to follow node.the js cli arguments convention
https://stackabuse.com/command-line-arguments-in-node-js/
```
yarn validator-logos --update-all 
yarn validator-logos --validator-address="inj...."
```

packages/sdk-ui-ts/validators-logo/index.ts
```diff
@@ -0,0 +1,256 @@
...
+        return { ...acc }
+      }
+
+      return { ...acc, [validatorAddress]: identity }
+    },
+    {} as ValidatorMap,
+  )
+}
+
+/* fetch url where validator image is stored */
```
ThomasRalee, line 64: Remove this comment

packages/sdk-ui-ts/validators-logo/index.ts
```diff
@@ -0,0 +1,256 @@
...
+const [, , flag] = process.argv
+const cliValidatorAddress = flag && flag.includes('inj') ? flag : undefined
+const shouldUpdateAllImages = flag && flag.includes('update:all')
+const removeImages = flag && flag.includes('remove:images')
+
+/* create validator to identity map */
+const fetchValidators = async () => {
+  const { validators } = await chainGrpcStakingApi.fetchValidators()
+
+  /* if validator is already in the map, don't include it by default because that would require redownloading all images */
```
ThomasRalee, line 33: Change this comment to:
`Filter out existing validator images`

packages/sdk-ui-ts/validators-logo/index.ts
```diff
@@ -0,0 +1,256 @@
...
+      const logoUrl = await fetchLogoUrl(identifier as string)
+
+      if (!logoUrl) {
+        return { ...(await acc) } as ValidatorMap
+      }
+
+      return { ...(await acc), [validatorAddress]: logoUrl }
+    },
+    {} as Promise<ValidatorMap>,
+  )
```
ThomasRalee, line 97: Lets use a map instead of reduce, we can do a Promise.all()

ThomasRalee (approved):

ThomasRalee (dismissed):

bangjelkoski (approved):

PR #178, opened 2023-04-27
chore: setup dev ci cd

ThomasRalee:


bootstrapEnv.js
```diff
@@ -0,0 +1,21 @@
+const fs = require('fs')
```
bangjelkoski, line 1: Let's move this file to the `etc` folder

.github/workflows/stable.yaml
```diff
@@ -0,0 +1,42 @@
+name: Publish stable
+
+on:
+  push:
+    branches:
+      - disabled
```
bangjelkoski, line 6: Why are you disabling it for now?

bangjelkoski (changes requested):

bangjelkoski (approved):

PR #179, opened 2023-04-29
chore: updateMetamaskNetwork helper function

shane-moore:


packages/wallet-ts/src/utils/wallets/metamask/Metamask.ts
```diff
@@ -17,3 +19,22 @@ export const getEthersProviderFromMetamask = async () => {
...
+      mustBeMetaMask: true,
+    })) as Eip1993ProviderWithMetamask
+
+    if (!provider) {
+      throw new WalletException(new Error('Please install Metamask Extension'))
+    }
+
+    await provider.request({
+      method: 'wallet_switchEthereumChain',
+      params: [{ chainId: `0x${chainId}` }],
```
bangjelkoski, line 35: This is not always true tho, it will work for us fine because we are always using `1` and `5` as chainIds, but let's say we want to switch to Kovan (or Arbitrum), Kovan is `42` and in hex its `2A`

bangjelkoski: Also, we should have a `return await` here

bangjelkoski (changes requested):

bangjelkoski (approved):

PR #207, opened 2023-06-29
Added STINJ token

peterwilli:
I added it as per https://github.com/InjectiveLabs/injective-ts/blob/master/packages/token-metadata/CONTRIBUTING.md

But I have no idea if I did it right, is there a way to test this?


packages/token-metadata/src/tokens/tokens/tokens.ts
```diff
@@ -562,6 +562,20 @@ export default {
...
 
+  STINJ: {
+    name: 'Stride Staked Injective',
+    symbol: 'STINJ',
+    decimals: 18,
+    logo: 'stinj.png',
+    coinGeckoId: 'stride-staked-injective',
+
+    cw20: {
+      decimals: 8,
```
bangjelkoski, line 573: Are you sure that the CW20 version has 8 decimals?

bangjelkoski (changes requested):

packages/token-metadata/src/tokens/tokens/tokens.ts
peterwilli: Honestly? Not sure. I copied it from the other cw20 files. I didn't know where to find it.

bangjelkoski: What're the decimals that the CW20 contract has? I think it should be 18, the same as the native INJ.

peterwilli: Where can I find this? I tried looking for it at: https://www.mintscan.io/injective/account/inj10fd06xl4q6jp9qlhemvm6ymmm83ppj2g8rzquw

peterwilli:
There's also aother potential thing we need to add, but I'm not sure how to get that either:

image

The stinj appears as a ibc token in the balance section in Injective Hub. But I'm really not sure where to find the right parameters to that (new to adding cosmos params). I found other contracts refering to IBC like [here](https://github.com/InjectiveLabs/injective-ts/blob/8241c1b143aa96fd6f961b3e4c75e0888c2c7998/packages/token-metadata/src/tokens/tokens/tokens.ts#L658).

peterwilli:
> Can you give me the denom of the stINJ?

Is that the full length of that IBC thing in the screenshot? If so, it says `ibc/AC87717EA002B0123B10A05063E69BCA274BA2C44D842AEEB41558D2856DCE93`

bangjelkoski:
```
chore: minor yarn (bangjelkoski)
@injectivelabs/bridge-ts: 1.12.0-beta.32
@injectivelabs/sdk-ts: 1.12.0-beta.29
@injectivelabs/sdk-ui-ts: 1.12.0-beta.30
@injectivelabs/token-metadata: 1.12.0-beta.17
@injectivelabs/wallet-ts: 1.12.0-beta.30
```

I've added the token meta. We'll update the products soon.

peterwilli:
Wow nice, thank you!

PR #222, opened 2023-08-15
chore: kava bridge

shane-moore:
## Changes
- typical setup for new IBC bridge - Kava
- refactored `canonicalChannelsToChainMapToInjective` to be an array of objects, since duplicate channel-Id's from foreign networks to Injective is possible. this map isn't used anywhere across the products, so was a little difficult to decide on the right data structure. if it becomes useful someday and another data structure is preferable, can modify then as well
- updated `canto` token's ibc hash because I noticed the previous hash was based off of `transfer/channel-99/canto` instead of `transfer/channel-99/acanto`


packages/token-metadata/src/ibc/channels.ts
```diff
@@ -59,24 +60,26 @@ export const canonicalChannelsToChainMapFromInjective = {
...
-  'channel-59': { chainA: 'Juno', chainB: 'Injective' },
-  'channel-6': { chainA: 'Stride', chainB: 'Injective' },
-  'channel-90': { chainA: 'Crescent', chainB: 'Injective' },
-  'channel-1': { chainA: 'Sommelier', chainB: 'Injective' },
-  'channel-8': { chainA: 'Canto', chainB: 'Injective' },
-  'channel-91': { chainA: 'Terra2', chainB: 'Injective' },
-  'channel-92': { chainA: 'Terra2', chainB: 'Injective' },
-  'channel-3': { chainA: 'Migaloo', chainB: 'Injective' },
-}
+export const canonicalChannelsToChainMapToInjective = [
```
bangjelkoski, line 66: Why don't we also add the above map here? `chainA` will always be injective in that case so we will have one array of objects instead of two. We just have to be careful to reflect these changes where these constants are used.

packages/token-metadata/src/tokens/tokens/tokens.ts
```diff
@@ -1685,4 +1685,20 @@ export default {
     tokenType: TokenType.TokenFactory,
   },
 
+  KAVA: {
+    name: 'KAVA',
+    symbol: 'KAVA',
+    decimals: 6,
+    logo: 'kava.webp',
+    coinGeckoId: 'kava',
+    ibc: {
```
bangjelkoski, line 1694: Blank row above

bangjelkoski (changes requested):

bangjelkoski (approved):

PR #226, opened 2023-08-21
Add Nebula token

Morgandri1:
Added as per `packages/token-metadata/CONTRIBUTING.md`


bangjelkoski:
I'll add it in our own commit now, the image was too big on this PR.

bangjelkoski:
Also, for this particular token factory denom you need to add the token metadata on chain: https://docs.ts.injective.network/core-modules/token-factory#msgsetdenommetadata

Morgandri1:
> Also, for this particular token factory denom you need to add the token metadata on chain: https://docs.ts.injective.network/core-modules/token-factory#msgsetdenommetadata

Do we need to re-do the token metadata then?

bangjelkoski:
I'll add it but for factory denoms it's best to use the on chain denom metadata because if someone adds another NBLA token they can overlap with yours so there will be a disrepancy.

Morgandri1:
I see, Looking at the 6 decimal spec i cant quite tell how to alter our `set-denom-metadata` command to reflect the required changes

bangjelkoski:
The example provided on the docs has 6 decimals :)

Morgandri1:
Yes, but we did this through the command line so im having some issues translating it. Apologies.

Morgandri1:
I was following [this](https://docs.injective.network/develop/guides/token_launch/) guide instead of the ts docs.

PR #228, opened 2023-08-23
feat: add logo for our validator

rayanebel:
The purpose of this PR is to add a logo for the validator: [injvaloper1dyntafvjynns5zyltwgyktej0y5zt84y3npkt4](https://www.mintscan.io/injective/validators/injvaloper1dyntafvjynns5zyltwgyktej0y5zt84y3npkt4)

PR #229, opened 2023-08-23
fix: wrong image name

rayanebel:
I sent a previous PR: https://github.com/InjectiveLabs/injective-ts/pull/228 but the filename was not correct.

This PR fix the image name for our validator.


rayanebel:
cc @bangjelkoski

PR #232, opened 2023-09-09
TienThuatToan validator logo

caotre:


bangjelkoski:
Hey, would it be a problem to provide a lower-size image? Maybe convert it to webp or just reduce the size, please?

PR #277, opened 2023-12-17
Update tokens.ts

followboat22:
token update


bangjelkoski:
Why have you removed the other tokens?

PR #290, opened 2023-12-18
adding ginger metadata

SHITcoinsonINJ:
Adding ginger metadata


bangjelkoski:
I'll deploy from the local branch, this PR is not good enough.

PR #298, opened 2023-12-19
ADD DOJ Metadata

injdev123:
add doj metadata including doj.png


bangjelkoski:
Will merge from local branch

PR #306, opened 2023-12-20
Update INJINU info

cryptodev-t:


bangjelkoski:
What on earth is going on with this token?! I'm removing it entirely now.

PR #328, opened 2023-12-23
Add native asset karate

zkevery:
Add token metadata for a native coin - karate


bangjelkoski:

PR #338, opened 2023-12-25
TOKEN INFO AND LOGO UPDATE

shibainu15:
TOKEN INFO AND LOGO UPDATE


bangjelkoski:
There is already one with that name.

PR #339, opened 2023-12-25
Update tokens.ts

moonbeamdoodles:
Added DUDE token


bangjelkoski:
Added from local branch

PR #341, opened 2023-12-26
Add Extra Virgin Olive Inu token data

EXTRAVIRGININU:
- Added token metadata
- Added token image


bangjelkoski:

PR #355, opened 2024-01-01
Update logo for SNOWY

InjSnowy:
logo for SNOWY token is updated.


bangjelkoski:

PR #364, opened 2024-01-05
Add updated data for waifu token

mitruth:


bangjelkoski:
I'll publish from local branch

PR #371, opened 2024-01-12
Refactor/token factory to token

bangjelkoski:
We've started having different tokens having the same symbol (mostly either multiple tokens having the same symbol as TokenFactory denoms or CW20 denoms). 

This PR adds a new `tokenFactory` (or `tokenFactories`) field which allows for having multiple token variations of the same symbol.


packages/sdk-ui-ts/src/services/ibc/ibcTokenMetadata.json
```diff
@@ -15,7 +15,8 @@
       "decimals": 18,
       "symbol": "INJ",
       "baseDenom": "INJ",
-      "isNative": false
+      "isNative": false,
```
ThomasRalee, line 18: Maybe we can take this opportunity to remove "isNative" boolean flag, dont seem to see any `isNative: true` instance

packages/token-metadata/src/TokenMetaUtils.ts
```diff
@@ -48,6 +48,40 @@ export class TokenMetaUtils {
...
+    if (tokenMeta.tokenFactory?.creator !== creatorAddress) {
+      return
+    }
+
+    return {
+      ...tokenMeta,
+      tokenType: TokenType.TokenFactory,
+      tokenVerification: TokenVerification.Verified,
+    }
+  }
```
ThomasRalee, line 83: Maybe we use this refactor opportunity to simplify the tokenmeta typing:
```
export interface TokenMeta {
  name: string
  logo: string
  symbol: string
  decimals: number
  tokenType: TokenType // primary token type
  tokenVerification?: TokenVerification
  coinGeckoId: string

  ibc?: IbcTokenMeta
  spl?: SplTokenMeta
  cw20?: Cw20TokenMeta
  cw20s?: Cw20TokenMetaWithSource[] // When there are multiple variations of the same CW20 token
  erc20?: Erc20TokenMeta
  evm?: EvmTokenMeta
  // tokenFactory?: NativeTokenFactoryMeta < === remove this
  tokenFactories?: NativeTokenFactoryMeta[] < === use this as the default
}
```

packages/token-metadata/src/utils.ts
```diff
@@ -21,14 +22,98 @@ import { TokenMetaUtilsFactory } from './TokenMetaUtilsFactory'
 const getCw20Meta = (
   token: Token,
 ): Cw20TokenMetaWithSource | Cw20TokenMeta | undefined => {
-  const denomToLowerCase = token.denom
+  const denomToLowerCase = token.denom.toLowerCase()
   const cw20MetaFromCw20s = token.cw20s?.find((meta) =>
```
ThomasRalee, line 26: should we update this to `token.tokenFactories?.find((meta) =>`

packages/token-metadata/src/TokenMetaUtils.ts
bangjelkoski: Yes, I was thinking the same, but the proper way to do it would be to have:

- One interface which is represented within the list `tokens.ts` which always has multiple `cw20s` and `tokenFactories` 
- One interface which we ship to the products which always has only one `cw20`

ThomasRalee: Would the UI implementation look something like this?
```
// in the store
getters: {
    getToken:(state) => (denomOrSymbol: string) {
        return (state.tokens as tokenMeta[]).find(...) as TokenMetaSingular
    }
}
```


```
// in the component 

token.cw20.decimal
token.tokenFactor.decimal 
```

bangjelkoski: Nope, the `TokenFactory.ts` class would handle this, particularly the `getTokenFromMeta` function which will always return one `cw20` and one `tokenFactory`. I'll do the refactor and push here so you guys can see

ThomasRalee (approved):

PR #377, opened 2024-01-16
add drogo

egrassy:
add drogo from terra


bangjelkoski:
Can you please reduce the size of the logo a bit? Or convert it to webp. Thanks

bangjelkoski (approved):

PR #381, opened 2024-01-22
$BITS token commit 

jadalien999:
Fixed the metadata format and added the image alongside in this commit.


bangjelkoski:
Token's structure is incorrect, you need to add the `tokenFactories` entry

bangjelkoski:
Added from a local branch.

PR #384, opened 2024-01-24
Added $IKINGS

InjectiveKings:
I hope we can merge :)


bangjelkoski:
Added from a local branch. Your image name was wrong in the token information, be careful next time.

PR #390, opened 2024-01-31
chore: perp markets token metadata

shane-moore:
## Changes
- For perp markets, we don't have a `denom` for the base symbol. We just know the `base symbol`, and we're trying to derive the token metadata from `token-metadata` package using `DenomClient.getDenomToken` in the same way that we would if we were trying to derive the metadata via passing the denom. This leads to edge cases and complexity
- In this PR, I propose that simply the flow by grabbing token metadata for derivatives markets "base tokens" via `DenomClient.getTokenMetaDataBySymbol` and then adding a few required fields to return a proper `Token` type. Note that it seems like the only useful fields for a derivative base token is the `name` / `symbol` / `logo` properties. The rest don't seem to be used, which is why I feel like the simplification I'm making here with the new `getDerivativesToken` could be justified

## Note
Placed this in draft because we should test some more before merging, but I'm curious your thoughts on this approach

### DIFF
--- a/packages/sdk-ui-ts/src/denom/DenomClientAsync.ts
+++ b/packages/sdk-ui-ts/src/denom/DenomClientAsync.ts
@@ -21,6 +21,7 @@ import {
 import { Web3Client } from '../services/web3/Web3Client'
 import type { Token } from '@injectivelabs/token-metadata'
 import {
+  TokenType,
   TokenMetaBase,
   getUnknownTokenWithSymbol,
   getIbcTokenFromDenomTrace,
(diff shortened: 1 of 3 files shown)


ThomasRalee:
The implementation looks fine at first glance:

It would be a lot easier to review if we had some test data to benchmark this implementation, i.e the before and after on the list of perp markets we currently support

bangjelkoski:
> The implementation looks fine at first glance:
> 
> It would be a lot easier to review if we had some test data to benchmark this implementation, i.e the before and after on the list of perp markets we currently support

I think that this is not the proper approach we should take. We should think where else are we using the symbol based search on the token-metadata and then think of an unified approach that'll take everything into consideration, we should not split the search based on each of the use cases

PR #394, opened 2024-02-07
Update validators logo

Grinion:
Update validators logo


bangjelkoski:
Have you used the script in `sdk-ui-ts` do to the update on the logos?

PR #396, opened 2024-02-14
fix: ibc metadata fetch error handling

shane-moore:


.github/workflows/dev.yaml
```diff
@@ -50,6 +50,9 @@ jobs:
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
+      - name: Get IBC Tokens
+        run: yarn --cwd packages/sdk-ui-ts get-ibc-tokens || true
```
shane-moore, line 54: the `|| true` will make the build continue even if error occurs when running `get-ibc-tokens` script

packages/sdk-ui-ts/src/services/ibc/tokens.ts
```diff
@@ -49,9 +49,24 @@ function ibcTokenMetadataToToken(
...
-    const response = (await ibcTokenMetadataApi.get(TOKEN_METADATA_PATH)) as {
+    const ibcTokenMetadataResponse = ibcTokenMetadataApi.get(
+      TOKEN_METADATA_PATH,
+    ) as Promise<{
+      data: IbcTokenMetadata[]
+    }>
+
+    const response = (await Promise.race([
+      ibcTokenMetadataResponse,
+      timeout(2000),
```
shane-moore, line 68: if response from ibc token metadata fetch takes longer than `2` seconds, the `timeout` function will throw an error, which will just get logged in the `catch` block below. then, build should continue

bangjelkoski: We can pass second parameter to HttpRestClient { timeout: 2000 } and have the same logic as you implemented

PR #408, opened 2024-03-04
Adding ODIN assets on INJ

AddySiz:


bangjelkoski:
I don't see token metadata information for `myrk` and `o9w`. Also, some of the images are too big in size, please reduce them or convert to `webp`

PR #458, opened 2024-04-08
refactor: token-metadata

ThomasRalee:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new token management system that allows for the retrieval of token metadata by symbol or denomination/address.
	- Added new token attributes to enhance token metadata information.

### DIFF
--- a/packages/token-metadata/src/TokenFactoryStatic.ts
+++ b/packages/token-metadata/src/TokenFactoryStatic.ts
@@ -1,0 +1,56 @@
+import { TokenType, TokenSource, TokenStatic } from './types'
+
+export class TokenFactoryStatic {
+  public registry: TokenStatic[]
+  public tokensByDenom: Record<string, TokenStatic>
+  public tokensBySymbol: Record<string, TokenStatic[]>
+
+  constructor(registry: TokenStatic[]) {
+    this.registry = registry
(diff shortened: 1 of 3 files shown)


packages/token-metadata/src/TokenFactoryStatic.ts
```diff
@@ -0,0 +1,56 @@
...
+        return list
+      }
+
+      return { ...list, [denom]: token }
+    }, {} as Record<string, TokenStatic>)
+
+    this.tokensBySymbol = registry.reduce((list, token) => {
+      const symbol = token.symbol.toLowerCase()
+
+      return { ...list, [symbol]: [...(list[symbol] || []), token] }
```
bangjelkoski, line 23: What will happen here if we have assets with same symbol but different denom?

bangjelkoski (changes requested):

packages/token-metadata/src/TokenFactoryStatic.ts
ThomasRalee: This map will return an array of Tokens: `Record<string, TokenStatic[]>`
reducing the list of tokens to searched based on symbol, for example, using source or type in the getMetaBySymbol function

PR #469, opened 2024-04-30
Fix Eip712TypedDataV2 for date-type

Grinion:
Hi, the **getEip712TypedDataV2** method doesn't work correctly for a Date. As a consequence, a transaction that has Messages with Date type fields cannot be correctly signed. For example for a message of type MsgGrant method generate:

Screenshot 2024-04-30 at 15 39 52

Blockchain expected:

Screenshot 2024-04-30 at 15 41 24

I'm not sure if this patch is good and won't break other Messages that use the Date type.  It may also be better to make this modification on the blockchain

## Summary by CodeRabbit

- **New Features**
	- Enhanced handling of Date objects in transaction data processing to improve accuracy and reliability.

### DIFF
--- a/packages/sdk-ts/src/core/modules/tx/eip712/eip712.ts
+++ b/packages/sdk-ts/src/core/modules/tx/eip712/eip712.ts
@@ -51,6 +51,13 @@ export const getEip712TypedData = ({
   }
 }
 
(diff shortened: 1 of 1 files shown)


bangjelkoski:
Is this PR still needed?

bangjelkoski:
nice one, will add in the next release

PR #527, opened 2024-12-07
chore: remove redundant words

guqicun:
## Summary by CodeRabbit

- **Documentation**
	- Enhanced descriptions for the `namedItem` method in both `HTMLAllCollection` and `HTMLFormControlsCollection` interfaces, clarifying the return types and conditions for multiple matches.


bangjelkoski:
These are automatically generated, and the changes will be overridden the next time we regenerate the API reference docs.

PR #540, opened 2025-02-10
fix: incorrect example titles in wallet-accounts.md

kerlos:
fix example title does not match with example code below


bangjelkoski:
Thanks, but we don't accept these kind of contributions usually done for airdrops.

kerlos:
I didn't do this for airdrop but i'm a developer currently working on injective related stuff.
please correct this some where in the future.

PR #573, opened 2025-06-30
feat: add support for dynamic gas fee

ThomasRalee:
## Summary by CodeRabbit

* **New Features**
  * Transaction fee calculations now use a dynamically fetched base fee from the blockchain, ensuring more accurate and up-to-date gas price estimates.
* **Bug Fixes**
  * Improved logic for fee estimation when accounts do not have a public key, resulting in more reliable transaction broadcasting in those scenarios.

### DIFF
--- a/packages/utils/src/helpers.ts
+++ b/packages/utils/src/helpers.ts
@@ -146,10 +146,10 @@ export const getStdFeeFromObject = (args?: {
           .toFixed(),
       },
     ],
+    gas: new BigNumberInBase(gasNormalized).toFixed(),
     payer /** for Web3Gateway fee delegation */,
     granter,
     feePayer,
-    gas: new BigNumberInBase(gasNormalized).toFixed(),
   }
 }
 
--- a/packages/utils/src/helpers.ts
+++ b/packages/utils/src/helpers.ts
@@ -3,8 +3,8 @@ import {
   DEFAULT_GAS_LIMIT,
   DEFAULT_GAS_PRICE,
 } from './constants.js'
-import BigNumberInBase from './classes/BigNumber/BigNumberInBase.js'
 import BigNumberInWei from './classes/BigNumber/BigNumberInWei.js'
+import BigNumberInBase from './classes/BigNumber/BigNumberInBase.js'
 import { Awaited } from './types.js'
 
 export const sleep = (timeout: number): Promise<void> =>
@@ -42,8 +42,8 @@ export const awaitForAll = async <T, S>(
 
 export const splitArrayToChunksThrow = <T>({
   array,
-  chunkSize,
   filter,
+  chunkSize,
 }: {
   array: Array<T>
   chunkSize: number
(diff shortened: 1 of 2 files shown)


packages/utils/src/helpers.ts
```diff
@@ -145,10 +146,10 @@ export const getStdFeeFromObject = (args?: {
           .toFixed(),
       },
     ],
-    gas: new BigNumberInBase(gasNormalized).toFixed(),
     payer /** for Web3Gateway fee delegation */,
     granter,
     feePayer,
+    gas: new BigNumberInBase(gasNormalized).toFixed(),
```
bangjelkoski: This change will break some wallets, the `signDoc` that we pass to Ledger using the legacy `amino` approach requires these to be sorted in the way that we have them right now.

packages/wallets/wallet-core/src/broadcaster/MsgBroadcaster.ts
```diff
@@ -1063,6 +1070,40 @@ export class MsgBroadcaster {
...
+      | string
+      | {
+          granter?: string
+          feePayer?: string
+          gasPrice?: string
+          gas?: string | number
+        },
+  ) {
+    const client = new ChainGrpcTxFeesApi(this.endpoints.grpc)
+    const baseFee = (await client.fetchEipBaseFee()).baseFee
```
bangjelkoski: We should fallback to the default one if the API call doesn't work

packages/wallets/wallet-core/src/broadcaster/MsgBroadcaster.ts
```diff
@@ -1063,6 +1070,40 @@ export class MsgBroadcaster {
...
+
+    if (!args) {
+      return getStdFee(baseFee ? { gasPrice: baseFee } : {})
+    }
+
+    if (typeof args === 'string') {
+      return getStdFee({
+        ...(baseFee && {
+          gasPrice: new BigNumberInBase(baseFee).plus(1).toFixed(),
```
bangjelkoski: whats up with the `plus(1)`?

bangjelkoski (changes requested):

packages/wallets/wallet-core/src/broadcaster/MsgBroadcaster.ts
```diff
@@ -1081,7 +1080,8 @@ export class MsgBroadcaster {
         },
   ) {
     const client = new ChainGrpcTxFeesApi(this.endpoints.grpc)
-    const baseFee = (await client.fetchEipBaseFee()).baseFee
+    const baseFee =
```
bangjelkoski: this is not good, what if the request fails and this throws an error? This while function will fail. There should be a try and catch somewhere here

bangjelkoski (changes requested):

bangjelkoski (approved):

PR #612, opened 2026-01-05
fix: allow passing grpc options into client classes

billyjacoby:
## Summary by CodeRabbit

* **New Features**
  * Support for additional gRPC web transport configuration options.

* **Improvements**
  * Lazy initialization of gRPC clients for reduced startup overhead and improved resource usage.
  * Centralized handling of transport metadata to streamline request behavior.

* **Chores**
  * Bumped package versions to 1.17.3-alpha.0.
  * Consolidated and refactored proto build scripts and packaging layout.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


packages/sdk-ts/src/utils/grpc.ts
```diff
@@ -39,4 +27,4 @@ export const getGrpcWebTransport = (
 /**
  * Re-export GrpcWebFetchTransport for direct use
  */
-export { GrpcWebFetchTransport }
+export { GrpcWebFetchTransport, type GrpcWebTransportAdditionalOptions }
```
ThomasRalee: Lets move type GrpcWebTransportAdditionalOptions into the type folder `/injective-ts/packages/sdk-ts/src/types/grpc.ts`

we should also re-export this from the `/injective-ts/packages/sdk-ts/src/types/index.ts` file

packages/sdk-ts/src/client/base/BaseGrpcConsumer.ts
```diff
@@ -17,19 +18,20 @@ export default class BaseGrpcConsumer {
...
     this.metadata = map
-    // Recreate transport with new metadata
+    // Recreate transport with new metadata, preserving existing options
     this.transport = new GrpcWebRpcTransport(this.endpoint, {
-      headers: this.metadata,
+      ...this.options,
+      meta: this.metadata,
     })
     return this
   }
```
ThomasRalee, line 45: Instead of passing options through super in every Class, it'll be simpler if we just reuse the setMetadata function here.

So the implementation would look something like this:

```
export type GrpcMetadataOptions = {
  metadata?: Record<string, string>
  options?: GrpcWebTransportAdditionalOptions
}

public setMetadata(metadataOptions: GrpcMetadataOptions) {
    this.options = metadataOptions.options
    this.metadata = metadataOptions.metadata
    // Recreate transport with new metadata, preserving existing options
    this.transport = new GrpcWebRpcTransport(this.endpoint, {
      meta: this.metadata,
      options: this.options,
    })

    return this
  }
  ```

```
export const indexerAuctionApi = new IndexerGrpcAuctionApi(
  ENDPOINTS.indexer
).setMetadata({
  metadata: { Referer: HTTP_REFERRER },
  options: ....
})
```

File: vitest.config.ts [hunk 3]
```diff
@@ -172,20 +172,20 @@ export default defineConfig({
...
         __dirname,
-        'protoV2/indexer/proto-ts/esm/generated',
+        'protoV2/indexer/proto-ts/generated',
       ),
       '@injectivelabs/indexer-proto-ts-v2/generated/injective_exchange_rpc_pb':
         join(
           __dirname,
-          'protoV2/indexer/proto-ts/esm/generated/injective_exchange_rpc_pb.mjs',
+          'protoV2/indexer/proto-ts/generated/injective_exchange_rpc_pb.js',
         ),
```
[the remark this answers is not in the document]
ThomasRalee: this is already fixed, can you verify?

PR #626, opened 2026-03-19
fix(eslint): preserve enum export and values in fixer

Ocheretovich:
avoid exporting non-exported enums and preserve original member values

## Summary by CodeRabbit

* **Bug Fixes**
  * Enum conversion now preserves export keywords so exported and non-exported enums are handled correctly.
  * Generated enum member values retain their original name casing rather than being lowercased.
  * Conversion now targets the correct declaration form for exported enums, ensuring replacements apply to the intended code region.

### DIFF
--- a/scripts/eslint-rules/no-enum.js
+++ b/scripts/eslint-rules/no-enum.js
@@ -30,14 +30,14 @@ export default {
         const parentNode = node.parent
         const isExported =
           parentNode && parentNode.type === 'ExportNamedDeclaration'
-        const exportKeyword = isExported ? '' : 'export '
+        const exportKeyword = isExported ? 'export ' : ''
 
         // Get the enum members
         const members = node.members.map((member) => {
           const key = member.id.name
           const value = member.initializer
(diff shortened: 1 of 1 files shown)


scripts/eslint-rules/no-enum.js
```diff
@@ -30,14 +30,14 @@ export default {
...
           parentNode && parentNode.type === 'ExportNamedDeclaration'
-        const exportKeyword = isExported ? '' : 'export '
+        const exportKeyword = isExported ? 'export ' : ''
 
         // Get the enum members
         const members = node.members.map((member) => {
           const key = member.id.name
           const value = member.initializer
             ? sourceCode.getText(member.initializer)
-            : `'${key.toLowerCase()}'`
```
ThomasRalee, line 40: Any reason why we want to remove `.toLowerCase()` ?

Ocheretovich: To preserve the original enum member value. Using toLowerCase() was changing it unexpectedly

PR #649, opened 2026-04-22
feat/rfq-executor-refinements

Frederick-88:
This pull request introduces several improvements and updates to the proto alignment workflow, SDK implementation, and testing conventions. The main themes are: enhanced proto audit and workflow documentation, improved SDK-side test coverage, and bugfixes/updates for the RFQ API and related modules.

**Proto Alignment Workflow Improvements**

* The proto alignment skill and workflow documentation now require a field-level audit for modified proto services, ensuring that changes are not assumed to be additive and that all message and RPC changes are mapped to SDK updates. This includes new guidance on using `awk` for message diffs and a table mapping proto changes to SDK actions. [[1]](diffhunk://#diff-0a5f9b584635de7e5e44d690286752549c0a45a364fe7713c39b5ff940e013d6L83-R105) [[2]](diffhunk://#diff-0a5f9b584635de7e5e44d690286752549c0a45a364fe7713c39b5ff940e013d6R126-R137) [[3]](diffhunk://#diff-3b5ba174d332baad4865216c7d45512eb10ffc136ce2eba5a892c6be44266179R79-R106) [[4]](diffhunk://#diff-3b5ba174d332baad4865216c7d45512eb10ffc136ce2eba5a892c6be44266179L91-R135)

* The workflow now clarifies handling of the "latest" version keyword, resolved via `npm view`, and details the approval and implementation steps, including file creation/editing and test file requirements. [[1]](diffhunk://#diff-0a5f9b584635de7e5e44d690286752549c0a45a364fe7713c39b5ff940e013d6L42-R43) [[2]](diffhunk://#diff-3b5ba174d332baad4865216c7d45512eb10ffc136ce2eba5a892c6be44266179L50-R50) [[3]](diffhunk://#diff-3b5ba174d332baad4865216c7d45512eb10ffc136ce2eba5a892c6be44266179L91-R135)

**SDK Test Coverage and Conventions**

* Test file conventions are formalized: one test per API method, using `Network.Mainnet` endpoints, type validation via `expect.objectContaining`, and specific patterns for stream API tests. Stream V2 test files are also required if streaming methods exist. [[1]](diffhunk://#diff-0a5f9b584635de7e5e44d690286752549c0a45a364fe7713c39b5ff940e013d6R362-R462) [[2]](diffhunk://#diff-0a5f9b584635de7e5e44d690286752549c0a45a364fe7713c39b5ff940e013d6R484-R485)

* New test file `IndexerGrpcRfqGwApi.spec.ts` is added, following the new conventions.

**RFQ API and Related Fixes**

* The `IndexerGrpcRFQApi` method signature and test are updated for correct field usage and order, ensuring all required fields are passed and mapped properly. [[1]](diffhunk://#diff-3c824675601f1157b8b72118346d1e426c1a36b0920f3663f1cc142c07eab70cR255-R298) [[2]](diffhunk://#diff-3c824675601f1157b8b72118346d1e426c1a36b0920f3663f1cc142c07eab70cR308-R314) [[3]](diffhunk://#diff-f008fe5cc58429d395a9b6c84c57c455b4dc888cf754d96c1bcb452d23205f59R119-L139)

* The `BaseGrpcConsumer` client method now accepts an optional `noRetry` flag to disable retries for certain calls.

**Dependency and Module Updates**

* The `@injectivelabs/indexer-proto-ts-v2` dependency is bumped to version `1.18.15`.

* The exceptions module map adds a new entry for `RfqGw`.

---

**Most Important Changes**

**Proto Alignment Workflow & Auditing**

- Added field-level audit for modified proto services, with detailed instructions and mapping of proto changes to SDK actions, improving reliability and maintainability of proto alignment. [[1]](diffhunk://#diff-0a5f9b584635de7e5e44d690286752549c0a45a364fe7713c39b5ff940e013d6L83-R105) [[2]](diffhunk://#diff-3b5ba174d332baad4865216c7d45512eb10ffc136ce2eba5a892c6be44266179R79-R106)
- Clarified version selection (supporting "latest" keyword) and approval/implementation steps in workflow documentation. [[1]](diffhunk://#diff-0a5f9b584635de7e5e44d690286752549c0a45a364fe7713c39b5ff940e013d6L42-R43) [[2]](diffhunk://#diff-3b5ba174d332baad4865216c7d45512eb10ffc136ce2eba5a892c6be44266179L50-R50) [[3]](diffhunk://#diff-3b5ba174d332baad4865216c7d45512eb10ffc136ce2eba5a892c6be44266179L91-R135)

**SDK Test Coverage**

- Formalized and expanded test file conventions for API and stream classes, and added a new test file for `IndexerGrpcRfqGwApi`. [[1]](diffhunk://#diff-0a5f9b584635de7e5e44d690286752549c0a45a364fe7713c39b5ff940e013d6R362-R462) [[2]](diffhunk://#diff-0a5f9b584635de7e5e44d690286752549c0a45a364fe7713c39b5ff940e013d6R484-R485) [[3]](diffhunk://#diff-9b37907e4e3187229b2b237ae69f1d468f42bbb4f5c97bed97992a1b6b8c580eR1-R39)

**RFQ API & Consumer Fixes**

- Updated `IndexerGrpcRFQApi` method and test for correct field handling and order, and added an optional `noRetry` flag to the base consumer. [[1]](diffhunk://#diff-3c824675601f1157b8b72118346d1e426c1a36b0920f3663f1cc142c07eab70cR255-R298) [[2]](diffhunk://#diff-3c824675601f1157b8b72118346d1e426c1a36b0920f3663f1cc142c07eab70cR308-R314) [[3]](diffhunk://#diff-f008fe5cc58429d395a9b6c84c57c455b4dc888cf754d96c1bcb452d23205f59R119-L139) [[4]](diffhunk://#diff-fada52eebb7abdad72b5d3df3509816fbbb0d2dc062014e84c5c83ee3e0de9fcR226-R235)

**Dependencies & Modules**

- Bumped `@injectivelabs/indexer-proto-ts-v2` to `1.18.15` and added `RfqGw` to the exceptions module map. [[1]](diffhunk://#diff-944e336933d1dfc3de0f34d191dace9ae82610bc0093397ccafeda2b6deab484L351-R351) [[2]](diffhunk://#diff-1bab200369d5d6f7619b9b2916dfeb2d6ef0933957b8655b7711674d01810ce5R46)

## New Features

* New RFQ Gateway API for preparing auto-signed orders with quote functionality
* Added time range filtering (startTime/endTime) to trade history queries
* Added ability to disable RPC retries on specific calls for improved control
* WebSocket streams now support conditional order acknowledgment and update events
* New transaction broadcast method supporting separate fee payer signatures


packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcRfqGwApi.ts
```diff
@@ -0,0 +1,156 @@
...
+      request.takerAddress = takerAddress
+    }
+
+    if (expiry) {
+      request.expiry = BigInt(expiry)
+    }
+
+    if (quotesWaitTimeMs) {
+      request.quotesWaitTimeMs = BigInt(quotesWaitTimeMs)
+    }
```
ivan-angjelkoski: ⚠️ **Potential issue — `fetchPrepareAutoSign` silently drops legitimate zero values**

Falsy guards skip the assignment when a numeric param is `0`. A freshly-granted auto-sign account's first broadcast has `sequence = 0`, and on some gateway configurations a new account's `accountNumber` can also be `0` — in both cases the prepared tx will be built with missing / incorrect auth info. Same class of bug for `expiry` and `quotesWaitTimeMs`.

Use explicit `!== undefined` checks for numeric fields (you already do this correctly for `subaccountNonce` at line 134).

```suggestion
    if (autosignAccountNumber !== undefined) {
      request.autosignAccountNumber = BigInt(autosignAccountNumber)
    }

    if (autosignAccountSequence !== undefined) {
      request.autosignAccountSequence = BigInt(autosignAccountSequence)
    }

    if (feePayerAccountNumber !== undefined) {
      request.feePayerAccountNumber = BigInt(feePayerAccountNumber)
    }

    if (feePayerAccountSequence !== undefined) {
      request.feePayerAccountSequence = BigInt(feePayerAccountSequence)
    }

    if (takerAddress) {
      request.takerAddress = takerAddress
    }

    if (expiry !== undefined) {
      request.expiry = BigInt(expiry)
    }

    if (quotesWaitTimeMs !== undefined) {
      request.quotesWaitTimeMs = BigInt(quotesWaitTimeMs)
    }
```

Also, `IndexerGrpcRfqGwApi.spec.ts` passes `autosignAccountNumber: 0` and `autosignAccountSequence: 0` — both are silently dropped today because the mainnet call is wrapped in try/catch, hiding the bug.

<details>

In `packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcRfqGwApi.ts` around lines 91 to 117, the guards `if (autosignAccountNumber)`, `if (autosignAccountSequence)`, `if (feePayerAccountNumber)`, `if (feePayerAccountSequence)`, `if (expiry)`, and `if (quotesWaitTimeMs)` silently drop legitimate zero values — a fresh auto-sign account's first tx has sequence `0`. Change each numeric guard to `!== undefined` so zero values are forwarded. Leave the string-typed `takerAddress` guard as-is. After fixing, update `IndexerGrpcRfqGwApi.spec.ts` to assert the zero values end up on the request (or document the behaviour explicitly).

</details>

packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcRfqGwApi.ts
```diff
@@ -0,0 +1,156 @@
...
+      request.worstPrice = worstPrice
+    }
+
+    if (autosignAddress) {
+      request.autosignAddress = autosignAddress
+    }
+
+    if (autosignPubKey) {
+      request.autosignPubKey = autosignPubKey
+    }
```
ivan-angjelkoski: 🧹 **Nitpick (non-blocking) — Required string fields use unnecessary falsy guards**

`clientId`, `marketId`, `direction`, `margin`, `quantity`, `worstPrice`, `autosignAddress`, and `autosignPubKey` are all typed as required (non-optional) in the param type. The `if (x) { ... }` wrappers add noise and blur which params are actually optional (`cid?`, `takerAddress?`, `expiry?`, …). If an empty string is ever passed, the caller is already violating the contract — assign directly and let the server reject.

```suggestion
    request.clientId = clientId
    request.marketId = marketId
    request.direction = direction
    request.margin = margin
    request.quantity = quantity
    request.worstPrice = worstPrice
    request.autosignAddress = autosignAddress
    request.autosignPubKey = autosignPubKey
```

<details>

In `packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcRfqGwApi.ts` around lines 59 to 89, the eight `if (x) { request.x = x }` guards wrap required string fields (`clientId`, `marketId`, `direction`, `margin`, `quantity`, `worstPrice`, `autosignAddress`, `autosignPubKey`). Replace each guarded block with a direct assignment so the code reflects the type contract. Leave guards for genuinely optional fields (`cid`, `takerAddress`, `unfilledAction`, etc.) unchanged.

</details>

packages/sdk-ts/src/client/indexer/ws/rfq/IndexerWsTakerStream.ts
```diff
@@ -177,6 +177,26 @@ export class IndexerWsTakerStream {
...
+
+        case 'conditional_order_update':
+          if (response.conditionalOrder) {
+            const order =
+              IndexerGrpcRfqTransformer.grpcConditionalOrderToConditionalOrder(
+                response.conditionalOrder,
+              )
+            this.emit('conditional_order_update', { order })
+          }
+          break
```
ivan-angjelkoski, line 198: 🔎 **Verification needed — `conditional_order_update` case reads `response.conditionalOrder`, while the ack case reads `response.conditionalOrderAck.order`**

These two arms use different payload shapes:

- `conditional_order_ack` → `response.conditionalOrderAck?.order` (wrapper + `order`)
- `conditional_order_update` → `response.conditionalOrder` (bare order)

This could be correct if the 1.18.15 proto exposes the update payload as a bare `ConditionalOrderResponseType` sibling, but asymmetric wrappers are a common source of subtle bugs and `node_modules` still has 1.18.14 locally so I can't verify directly. Please confirm against the regenerated `injective_rfq_rpc_pb.d.ts` after `pnpm install` — if `TakerStreamResponse` defines the update as e.g. `conditionalOrderUpdate?: ConditionalOrderUpdate { order }`, this should read `response.conditionalOrderUpdate?.order`.

<details>

In `packages/sdk-ts/src/client/indexer/ws/rfq/IndexerWsTakerStream.ts` around lines 190 to 198, verify against the regenerated `@injectivelabs/indexer-proto-ts-v2@1.18.15` file `generated/injective_rfq_rpc_pb.d.ts`: inspect the `TakerStreamResponse` interface for the field backing the `conditional_order_update` messageType. If the proto defines a wrapper (e.g. `conditionalOrderUpdate?: ConditionalOrderUpdate { order: ConditionalOrderResponseType }`), change the handler to read `response.conditionalOrderUpdate?.order` to match the symmetric shape of the `conditional_order_ack` case. If it truly is a bare `conditionalOrder?: ConditionalOrderResponseType`, leave the code and add a one-line comment noting the asymmetry is intentional.

</details>

packages/sdk-ts/src/client/indexer/transformers/IndexerTcDerivativesStreamTransformer.ts
```diff
@@ -11,18 +11,15 @@ export class IndexerTcDerivativesStreamTransformer {
...
-            order,
-          )
-        : undefined,
+      order:
+        order && order.state !== OrderState.Booked
+          ? IndexerGrpcTcDerivativesTransformer.grpcOrderHistoryToOrderHistory(
+              order,
+            )
+          : undefined,
     }
```
ivan-angjelkoski, line 26: ⚠️ **Potential issue — Booked-state filtering changed from "drop event entirely" to "emit with `order: undefined`"**

Previously this callback returned `undefined` when the order state was `Booked`, which meant subscribers never observed the event. After this change the callback emits `{ operationType, timestamp, order: undefined }` — downstream consumers that only listened for "real" order updates will now see these as empty-order events, and any `if (event.order)` gating in callers will silently suppress them.

If this behaviour change is intentional (e.g. you want consumers to see the `operationType` / `timestamp` even for Booked), please call it out in the PR description. If not, restore the early return:

```suggestion
    return {
      operationType: response.operationType,
      timestamp: Number(response.timestamp),
      order:
        order && order.state !== OrderState.Booked
          ? IndexerGrpcTcDerivativesTransformer.grpcOrderHistoryToOrderHistory(
              order,
            )
          : undefined,
    }
```

(Replace with the suggestion above only if you want to keep the new behaviour and merely document it. Otherwise, reintroduce `if (order?.state === OrderState.Booked) return` above the `return` statement and revert the `order:` ternary to its pre-refactor form.)

<details>

In `packages/sdk-ts/src/client/indexer/transformers/IndexerTcDerivativesStreamTransformer.ts` around lines 14 to 23, the `orderHistoryStreamCallback` previously early-returned when `order.state === OrderState.Booked`, dropping the event entirely. The refactor now emits `{ operationType, timestamp, order: undefined }` for Booked orders, which is a semantic change. If the intent was only to avoid transforming Booked orders (not to drop the event), keep the new shape and document it in the PR body. Otherwise restore the pre-refactor early-return (`if (order?.state === OrderState.Booked) return`) above the `return` block so Booked events remain fully suppressed.

</details>

packages/sdk-ts/src/core/tx/broadcaster/MsgBroadcasterWithPk.ts
```diff
@@ -442,6 +446,48 @@ export class MsgBroadcasterWithPk {
...
+    })
+
+    const signDocBytes = CosmosTxV1Beta1TxPb.SignDoc.toBinary(signDoc)
+    const signature = await privateKey.sign(signDocBytes)
+
+    const feePayerSigBytes = feePayerSig.startsWith('0x')
+      ? hexToUint8Array(feePayerSig.slice(2))
+      : base64ToUint8Array(feePayerSig)
+
+    txRaw.signatures = [signature, feePayerSigBytes]
```
ivan-angjelkoski, line 486: 🛠️ **Refactor suggestion — Assert `AuthInfo` has two SignerInfos before overwriting `signatures`**

`txRaw.signatures = [signature, feePayerSigBytes]` only yields a valid tx when the incoming `AuthInfo` already contains two `SignerInfos` in `[signer, feePayer]` order. The tx comes from the RFQ gateway so this should be the case in practice, but a single misconfigured field on the server side would silently produce an invalid tx that fails on-chain with a confusing error.

A cheap sanity check fails fast and documents the contract:

```suggestion
    if (txRaw.signatures.length !== 2) {
      throw new GeneralException(
        new Error(
          'broadcastWithFeePayerSig expects tx with exactly 2 SignerInfos (signer, feePayer), got ' +
            txRaw.signatures.length,
        ),
      )
    }

    txRaw.signatures = [signature, feePayerSigBytes]
```

<details>

In `packages/sdk-ts/src/core/tx/broadcaster/MsgBroadcasterWithPk.ts` around line 486, the assignment `txRaw.signatures = [signature, feePayerSigBytes]` silently assumes the incoming `AuthInfo` already has two SignerInfos in `[signer, feePayer]` order. Before the assignment, add a guard that throws a `GeneralException` when `txRaw.signatures.length !== 2`, so a misconfigured gateway response fails fast instead of producing a malformed on-chain tx.

</details>

ivan-angjelkoski (commented):
## Review summary

Seven findings, one critical:

- **Critical** — `fetchPrepareAutoSign` silently drops zero values for auth-related numeric params; fresh auto-sign accounts will broadcast with missing sequence / account number.
- **Medium** — Verify `conditional_order_update` proto field shape against the regenerated `1.18.15` types; `broadcastWithFeePayerSig` lacks a SignerInfo-count guard; `IndexerTcDerivativesStreamTransformer` silently changed Booked-event filtering from "drop event" to "emit with empty order".
- **Minor** — Redundant falsy guards on required string params; `RFQConditionalOrder.error` could be optional; `SKILL.md` checklist contradicts the actual export-ordering convention.

Positives: new test-file conventions are cleanly applied across three new spec files, `noRetry` on `executeGrpcCall` is a clean addition, `createConditionalOrder` field reordering matches the proto, and the Step 7a field-level audit codifies exactly the class of bug this PR catches (new `messageType` needing a WS case arm).

packages/sdk-ts/src/client/indexer/types/rfq.ts
```diff
@@ -147,18 +147,20 @@ export interface RFQConditionalOrderInput {
 
 export interface RFQConditionalOrder {
   rfqId: number
-  marketId: string
-  direction: string
+  error: string
```
ivan-angjelkoski: 🧹 **Nitpick (non-blocking) — `error` on a healthy order will be `""`; mark as optional**

Required-typed fields signal "this is always meaningful"; in practice `error` will be `""` for every non-failed order. Making it optional (`error?: string`) matches intent and prevents downstream `if (order.error)` checks from looking defensively-unused.

```suggestion
  error?: string
```

<details>

In `packages/sdk-ts/src/client/indexer/types/rfq.ts` around line 150, the `RFQConditionalOrder.error` field is typed `error: string`. Change it to `error?: string` to reflect that healthy (non-errored) orders have no meaningful error value. Verify callers and the transformer map the field unconditionally; if they do, they are still correct because the proto returns empty string which narrows fine.

</details>

.claude/skills/align-proto/SKILL.md
```diff
@@ -352,6 +481,8 @@ When implementing a new service, add the module to the exceptions package and bu
 - [ ] JSDoc comments match existing style: `/** @category */`
 - [ ] Alphabetical ordering in index.ts exports
```
ivan-angjelkoski: 📝 **Documentation nit — Checklist contradicts the actual convention**

`packages/sdk-ts/src/client/indexer/grpc/index.ts` and `transformers/index.ts` both sort exports by **line length** (visual pyramid), not alphabetically — the new `IndexerGrpcRfqGwApi` / `IndexerGrpcRfqGwTransformer` entries were inserted correctly by length. Either update the checklist wording to match the convention actually used, or reflow the index files.

```suggestion
- [ ] Ordering in index.ts exports follows existing convention (length-sorted pyramid)
```

<details>

In `.claude/skills/align-proto/SKILL.md` around line 482, the checklist item reads `- [ ] Alphabetical ordering in index.ts exports` but `packages/sdk-ts/src/client/indexer/grpc/index.ts` and `transformers/index.ts` actually sort by line length (visual pyramid). Either update the checklist to describe the length-sorted convention, or reformat every affected `index.ts` file to be truly alphabetical. If the latter, also audit other sdk-ts index files that currently follow the same pyramid style.

</details>

packages/sdk-ts/src/client/indexer/transformers/IndexerTcDerivativesStreamTransformer.ts
ivan-angjelkoski, line 26: ⚠️ **Potential issue — Booked orders now emit `order: undefined` instead of being dropped**

Previously, when `response.order.state === OrderState.Booked`, the callback returned early (`return`) and the entire stream event was suppressed — downstream subscribers never saw it. After this refactor, the callback always returns an object: when the order is `Booked` it now emits `{ operationType, timestamp, order: undefined }`.

This is a silent semantic shift for every existing consumer of `orderHistoryStreamCallback`:

- Subscribers that assumed every emission corresponds to a non-Booked order will now receive `order: undefined` events they previously never saw.
- Any code doing `event.order.id` (without an `undefined` guard) will start crashing.
- Any code counting/aggregating events (e.g. "order updates per second") will see inflated counts.
- The UI may flash "unknown order" rows for what were previously silent Booked emissions.

This is **not** equivalent to the previous behavior — the old code dropped the *whole stream event*, while the new code drops only the `order` payload but still propagates `operationType` + `timestamp`. If the intent really is to suppress Booked-state events as before, restore the early `return`.

<details>

```suggestion
    const order = response.order

    if (order?.state === OrderState.Booked) {
      return
    }

    return {
      operationType: response.operationType,
      timestamp: Number(response.timestamp),
      order: order
        ? IndexerGrpcTcDerivativesTransformer.grpcOrderHistoryToOrderHistory(
            order,
          )
        : undefined,
    }
```

</details>

<details>

In `packages/sdk-ts/src/client/indexer/transformers/IndexerTcDerivativesStreamTransformer.ts` around lines 11-24, the previous behavior dropped the whole stream event when `order.state === OrderState.Booked` by returning early; the new code instead always returns an event object with `order: undefined`, which changes downstream emission semantics. Restore the original early `return` when `order?.state === OrderState.Booked` so Booked events are suppressed entirely (no `{ operationType, timestamp, order: undefined }` emission), and keep the simple `order ? transform(order) : undefined` mapping for the non-Booked path. If the new behavior is intentional, document the migration and audit subscribers for `order: undefined` handling.

</details>

packages/sdk-ts/src/client/indexer/types/rfq.ts
```diff
@@ -147,18 +147,20 @@ export interface RFQConditionalOrderInput {
...
   createdAt: number
   updatedAt: number
   expiresAt: number
+  eventTime: number
+  worstPrice: string
   triggerType: string
+  triggerPrice: string
+  requestAddress: string
   minTotalFillQuantity: string
 }
```
ivan-angjelkoski, line 165: 🛠️ **Refactor suggestion — `RFQConditionalOrder` gained a required `eventTime` field**

`RFQConditionalOrder` is a public re-exported interface (see `types/index.ts`). The diff adds a new **required** field `eventTime: number` (and a new optional `error?: string`).

Any external consumer that builds an `RFQConditionalOrder` literal — tests, mocks, derived adapters, application-layer factories — will fail to type-check until they add an `eventTime` value. There is no migration note in the PR description and the field has no default.

Also note that `IndexerGrpcRfqTransformer.grpcConditionalOrderToConditionalOrder` reads `Number(grpcOrder.eventTime)` — when older proto responses (≤ 1.18.14) lack this field, `Number(undefined)` yields `NaN`, which then propagates into apps. Either make `eventTime?: number` (optional) or coerce missing values to `0` in the transformer (see related comment).

<details>

```suggestion
  rfqId: number
  error?: string
  margin: string
  status: string
  marketId: string
  quantity: string
  direction: string
  createdAt: number
  updatedAt: number
  expiresAt: number
  eventTime?: number
  worstPrice: string
  triggerType: string
  triggerPrice: string
  requestAddress: string
  minTotalFillQuantity: string
```

</details>

<details>

In `packages/sdk-ts/src/client/indexer/types/rfq.ts` around lines 147-166, the public interface `RFQConditionalOrder` added `eventTime: number` as a required field, which is a breaking change for consumers constructing this type. Either (a) make `eventTime?: number` to preserve backwards compatibility, or (b) keep it required but document the migration in the PR/CHANGELOG and bump the package version. Also update `IndexerGrpcRfqTransformer.grpcConditionalOrderToConditionalOrder` to coerce missing values via `Number(grpcOrder.eventTime ?? 0)` so older proto responses don't silently produce `NaN`.

</details>

packages/sdk-ts/src/client/indexer/transformers/IndexerGrpcRfqTransformer.ts
```diff
@@ -150,30 +150,32 @@ export class IndexerGrpcRfqTransformer {
...
       triggerPrice: grpcOrder.triggerPrice,
-      status: grpcOrder.status,
       createdAt: Number(grpcOrder.createdAt),
       updatedAt: Number(grpcOrder.updatedAt),
       expiresAt: Number(grpcOrder.expiresAt),
-      triggerType: grpcOrder.triggerType,
+      eventTime: Number(grpcOrder.eventTime),
+      requestAddress: grpcOrder.requestAddress,
       minTotalFillQuantity: grpcOrder.minTotalFillQuantity,
     }
```
ivan-angjelkoski, line 180: 🛠️ **Refactor suggestion — `Number(grpcOrder.eventTime)` produces `NaN` when the proto field is absent**

`grpcConditionalOrderToConditionalOrder` now sets `eventTime: Number(grpcOrder.eventTime)`. The `event_time` field was added in `indexer-proto-ts-v2@1.18.15` (this PR bumps the dep), but consumers running against older backends — or backends rolling out gradually — may emit messages where the field is absent. `Number(undefined)` is `NaN`, which then leaks through the public `RFQConditionalOrder` to UI/state code.

Same `NaN` risk applies to `createdAt`/`updatedAt`/`expiresAt` if those are missing, but they were not changed here; `eventTime` is the new field most likely to be missing in the wild.

Guard the conversion to default to `0`:

<details>

```suggestion
      eventTime: Number(grpcOrder.eventTime ?? 0),
```

</details>

<details>

In `packages/sdk-ts/src/client/indexer/transformers/IndexerGrpcRfqTransformer.ts` around lines 161-181, `eventTime: Number(grpcOrder.eventTime)` will produce `NaN` when the proto response omits the new `event_time` field (e.g. older backends). Coalesce to `0` with `Number(grpcOrder.eventTime ?? 0)` to keep the public `RFQConditionalOrder` shape numeric, and consider doing the same review for any other newly-added uint64 fields read here.

</details>

packages/sdk-ts/src/client/indexer/ws/rfq/IndexerWsTakerStream.ts
```diff
@@ -177,6 +177,26 @@ export class IndexerWsTakerStream {
...
+
+        case 'conditional_order_update':
+          if (response.conditionalOrder) {
+            const order =
+              IndexerGrpcRfqTransformer.grpcConditionalOrderToConditionalOrder(
+                response.conditionalOrder,
+              )
+            this.emit('conditional_order_update', { order })
+          }
```
ivan-angjelkoski, line 197: 🛠️ **Refactor suggestion — `conditional_order_update` silently swallowed when payload missing**

The new `conditional_order_update` arm reads `response.conditionalOrder` and only emits when truthy. Two concerns:

1. **Silent drop on missing payload:** when `response.conditionalOrder` is falsy, the case `break`s with no log/warn/error emission — subscribers will never know an update was missed. The neighboring `error` case at least surfaces an error; here a malformed update is silently swallowed. The same applies to `conditional_order_ack` reading `response.conditionalOrderAck?.order`: an ack arriving without an `order` payload (e.g. a receipt-only ack) is silently dropped, and any caller waiting on the ack will hang indefinitely.

2. **Field-name asymmetry:** `conditional_order_ack` reads `response.conditionalOrderAck?.order` (a wrapper containing `.order`), while `conditional_order_update` reads `response.conditionalOrder` directly. Please double-check the proto for `StreamTakerEventsResponse`: if the update payload is also wrapped (e.g. `response.conditionalOrderUpdate.order`), this branch will silently no-op — the truthy guard treats the wrapper as truthy and passes an object that lacks the conditional-order fields, so the transformer reads `undefined` everywhere and `Number(undefined) = NaN`.

<details>

```suggestion
        case 'conditional_order_update':
          if (response.conditionalOrder) {
            const order =
              IndexerGrpcRfqTransformer.grpcConditionalOrderToConditionalOrder(
                response.conditionalOrder,
              )
            this.emit('conditional_order_update', { order })
          } else {
            this.emit('error', {
              code: -1,
              message:
                'conditional_order_update received without conditionalOrder payload',
            })
          }
          break
```

</details>

<details>

In `packages/sdk-ts/src/client/indexer/ws/rfq/IndexerWsTakerStream.ts` around lines 180-197, the new `conditional_order_ack` and `conditional_order_update` cases silently drop the event when the expected payload field is missing (`response.conditionalOrderAck?.order` or `response.conditionalOrder`). Add an `else` branch that surfaces the malformed message — either emit an `error` stream event or log a warning — so subscribers waiting on an ack don't hang. Also verify against the proto whether `conditional_order_update` actually carries the order at `response.conditionalOrder` directly or under a wrapper like `response.conditionalOrderUpdate.order`; align the access with the schema.

</details>

ivan-angjelkoski (commented):
## Review summary

Ten findings across four lenses (general / regression / security / performance), one **critical**:

- **Critical** — `IndexerTcDerivativesStreamTransformer.orderHistoryStreamCallback` no longer drops the whole stream event for `Booked` orders; it now emits `{ operationType, timestamp, order: undefined }`. This is a silent semantic shift for every existing subscriber.
- **Medium (regression)** — `RFQConditionalOrder` gained a **required** `eventTime: number` field; `Number(grpcOrder.eventTime)` produces `NaN` when older backends omit the field; new WS arms `conditional_order_ack` / `conditional_order_update` silently swallow events with missing payloads (and the field-access path differs between the two arms — verify against proto).
- **Medium (security)** — `broadcastWithFeePayerSig` signs caller-supplied tx bytes without validating message contents, signer-layout (`AuthInfo.signerInfos[0].publicKey` vs local key), or `feePayerSig` length; encoding is inferred from a `0x` prefix with no length sanity check.
- **Minor** — `createConditionalOrder` spec scaled `deadlineMs` by 1000 (verify ms-vs-s unit shift); redundant `!== null` guards on `number | undefined` params in `IndexerGrpcTcDerivativesApi`; `rfq-gw.ts` imports a sibling type via the barrel (circular); `BaseGrpcConsumer.executeGrpcCall(noRetry)` needs JSDoc and a check that `retry()` carries no auth-refresh side-effects.

**Performance:** no findings. Transformers are pure object construction, the new gRPC call is a single non-retried RPC, and no unbounded loops/caches/subscriptions were introduced.

**Positives:**

- The new test-file conventions are cleanly applied — `IndexerGrpcRfqGwApi.spec.ts` and `IndexerGrpcTcDerivativesApi.spec.ts` follow the documented pattern (one test per method, `Network.Mainnet`, `expect.objectContaining`).
- `noRetry` on `executeGrpcCall` is a clean, opt-in addition that preserves the existing default behavior for all current call-sites.
- The proto-alignment workflow docs now explicitly require a field-level audit for modified services rather than assuming additive changes — a meaningful improvement to PR-review reliability.
- The new `IndexerGrpcRfqGwApi` client and transformer follow the existing module layout, including matching naming, JSDoc, and barrel-export style (modulo the one circular-import nit).

packages/sdk-ts/src/core/tx/broadcaster/MsgBroadcasterWithPk.ts
```diff
@@ -442,6 +446,48 @@ export class MsgBroadcasterWithPk {
...
+
+    return await this.broadcastTxRaw(txRaw)
+  }
```
ivan-angjelkoski, line 489: 🔎 **Verification needed — `broadcastWithFeePayerSig` signs caller-supplied tx bytes without validating contents or signer layout**

This new method takes pre-built `tx` bytes (from the RFQ gateway response), deserializes them via `TxRaw.fromBinary`, and unconditionally signs the resulting `SignDoc` with `this.privateKey`. The decoded `bodyBytes` / `authInfoBytes` are passed straight through — the method never inspects:

- **Message contents** (`bodyBytes` could decode to a `MsgSend`, `MsgExecuteContract`, or any other tx that drains the user's funds — the caller has full control over what the user signs).
- **Signer layout in `AuthInfo`**: `txRaw.signatures = [signature, feePayerSigBytes]` assumes exactly two `SignerInfo` entries with the user at index 0 and the fee-payer at index 1. If the gateway returns an `AuthInfo` whose `SignerInfos[0].publicKey` is not derived from `this.privateKey`, the user signs a doc whose first signer-slot is somebody else.
- **Fee / fee_payer fields** in the decoded `AuthInfo` (caller can set arbitrary fee, gas limit, granter, etc.).

Unlike the normal `broadcast` path — which builds the tx itself via `prepareTxForBroadcast` — this method essentially turns the SDK into a blind signing oracle for whatever bytes the RFQ gateway (or anyone able to MITM the gateway endpoint) returns. The existing OFAC check at line ~458 only blocks blacklisted private keys; it does not constrain message contents.

Please confirm the threat model assumes the RFQ gateway is fully trusted **and** that the gRPC channel uses TLS / cert-pinning. If so, document the assumption inline. Otherwise add at minimum:

1. Decode `txRaw.authInfoBytes` → `AuthInfo` and assert exactly two `SignerInfos`, with `signerInfos[0].publicKey` matching `privateKey.toPublicKey()`.
2. Decode `txRaw.bodyBytes` → `TxBody` and assert `messages` are restricted to the expected RFQ message types.
3. Bound `feePayerSig` length after decoding (typically 64 or 65 bytes).

The wallet-core variant (`packages/wallets/wallet-core/src/broadcaster/MsgBroadcaster.ts`) should mirror whatever decision is made here.

<details>

In `packages/sdk-ts/src/core/tx/broadcaster/MsgBroadcasterWithPk.ts` around lines 449 to 489, `broadcastWithFeePayerSig` signs caller-supplied tx bytes without validating their contents or signer layout. Before producing the signature, decode `txRaw.authInfoBytes` into an `AuthInfo` and require exactly two `SignerInfos`; assert that `signerInfos[0].publicKey` matches the public key derived from `this.privateKey` (so the user is provably signing for their own slot). Optionally decode `txRaw.bodyBytes` into a `TxBody` and verify the messages are restricted to the expected RFQ message types (reject anything else with `GeneralException`). Also bound the length of `feePayerSig` before decoding. Mirror the same hardening in `packages/wallets/wallet-core/src/broadcaster/MsgBroadcaster.ts`. If the design intentionally treats the RFQ gateway as fully trusted, document that assumption with an inline comment and ensure the gRPC transport enforces TLS.

</details>

packages/sdk-ts/src/core/tx/broadcaster/MsgBroadcasterWithPk.ts
```diff
@@ -442,6 +446,48 @@ export class MsgBroadcasterWithPk {
...
+    const signDocBytes = CosmosTxV1Beta1TxPb.SignDoc.toBinary(signDoc)
+    const signature = await privateKey.sign(signDocBytes)
+
+    const feePayerSigBytes = feePayerSig.startsWith('0x')
+      ? hexToUint8Array(feePayerSig.slice(2))
+      : base64ToUint8Array(feePayerSig)
+
+    txRaw.signatures = [signature, feePayerSigBytes]
+
+    return await this.broadcastTxRaw(txRaw)
```
ivan-angjelkoski, line 488: 🛠️ **Refactor suggestion — `feePayerSig` encoding inferred from `0x` prefix without length validation**

`broadcastWithFeePayerSig` infers the encoding of `feePayerSig` purely from a `0x` prefix: anything starting with `0x` is treated as hex, everything else is treated as base64. Two concerns:

1. A hex string **without** the `0x` prefix (a common server convention) will be silently base64-decoded, producing garbage signature bytes and a confusing on-chain rejection rather than a clear client-side error.
2. There is no validation that the decoded length is a sensible signature length (typically 64 or 65 bytes), so a malformed input fails far from the source.

Document the expected format on the parameter (JSDoc on `feePayerSig`) and add a post-decode length check that throws a clear `GeneralException` if the byte length is unexpected.

<details>

In `packages/sdk-ts/src/core/tx/broadcaster/MsgBroadcasterWithPk.ts` around lines 484-490, `broadcastWithFeePayerSig` decides between hex and base64 decoding for `feePayerSig` purely based on a `0x` prefix, with no validation. Add a JSDoc on the parameter clarifying the accepted encodings, and add a post-decode length sanity check (64 or 65 bytes) that throws a clear `GeneralException` if the decoded signature has an unexpected length, so misencoded inputs fail with a useful error instead of producing an invalid broadcast.

</details>

packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcRfqApi.ts
```diff
@@ -252,54 +252,50 @@ export class IndexerGrpcRFQApi extends BaseIndexerGrpcConsumer {
...
+      laneVersion: bigint
       triggerPrice: string
-      unfilledAction?: string
-      cid?: string
+      contractAddress: string
+      subaccountNonce: number
       allowedRelayer?: string
+      unfilledAction?: string
+      minTotalFillQuantity: string
     }
```
ivan-angjelkoski, line 277: 🔎 **Verification needed — `createConditionalOrder` spec scaled `deadlineMs` by 1000**

The destructured parameter type for `createConditionalOrder` was reordered (`signature` moved to top, `unfilledAction` block moved below `cid` / `allowedRelayer`). For TypeScript callers passing object literals, reordering is a no-op — the structural type is unchanged.

However, the spec test (`IndexerGrpcRfqApi.spec.ts`) materially changed its sample values. In particular:

- `deadlineMs: BigInt(1717000000)` → `BigInt(1717000000000)` (×1000)
- `quantity: '1'` → `'1000000000000000000'`
- `epoch`, `laneVersion`, `rfqId`, `minTotalFillQuantity` all changed

The `deadlineMs` change suggests the backend now interprets the field as **milliseconds since epoch** rather than seconds. If that's the case, every existing caller passing `deadlineMs` in seconds is now silently submitting an order with a deadline ~1970, which the backend may either reject or accept as already-expired.

Please confirm the unit semantics of `deadlineMs`. If they shifted, add a JSDoc on the field clarifying it's milliseconds and call this out in the PR description / CHANGELOG as a behavior change for existing callers.

<details>

In `packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcRfqApi.ts` around lines 252-301 (`createConditionalOrder` parameter type), the field set is unchanged but the corresponding spec test was rewritten with `deadlineMs` scaled by 1000, suggesting the backend now expects milliseconds since epoch instead of seconds. Confirm the unit with the proto/backend owners; if the semantics shifted, add a JSDoc comment on `deadlineMs` clarifying it is milliseconds, and call this out in the PR description as a behavior change for existing callers.

</details>

packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcTcDerivativesApi.ts
```diff
@@ -104,6 +108,14 @@ export class IndexerGrpcTcDerivativesApi extends BaseIndexerGrpcConsumer {
       request.sortDirection = sortDirection
     }
 
+    if (startTime !== undefined && startTime !== null) {
+      request.startTime = BigInt(startTime)
+    }
+
+    if (endTime !== undefined && endTime !== null) {
+      request.endTime = BigInt(endTime)
+    }
```
ivan-angjelkoski: 🧹 **Nitpick (non-blocking) — Redundant `!== null` guard on `number | undefined` params**

`startTime` and `endTime` are typed as `number | undefined` (no `null`), so the `&& x !== null` halves are dead code. Other guards in this file use the simpler `if (token)` / `if (sortBy)` pattern. The current check correctly handles `0` (because of the `!== undefined` half), so there is no correctness issue — only style/consistency.

<details>

```suggestion
    if (startTime !== undefined) {
      request.startTime = BigInt(startTime)
    }

    if (endTime !== undefined) {
      request.endTime = BigInt(endTime)
    }
```

</details>

<details>

In `packages/sdk-ts/src/client/indexer/grpc/IndexerGrpcTcDerivativesApi.ts` around lines 111-117, the `startTime`/`endTime` guards include a redundant `!== null` check for parameters typed as `number | undefined`. Simplify each guard to `if (startTime !== undefined)` / `if (endTime !== undefined)` to match the style used elsewhere in this file.

</details>

packages/sdk-ts/src/client/indexer/types/rfq-gw.ts
```diff
@@ -0,0 +1,59 @@
+import type * as InjectiveRfqGwRpcPb from '@injectivelabs/indexer-proto-ts-v2/generated/injective_rfq_gw_rpc_pb'
+import type { RFQSettlementUnfilledActionType } from '../types'
```
ivan-angjelkoski: 🧹 **Nitpick (non-blocking) — Self-referential import via barrel file**

`rfq-gw.ts` imports `RFQSettlementUnfilledActionType` from `'../types'` (the barrel `types/index.ts`), which in turn re-exports `rfq-gw.ts` itself. This creates a circular import path between the new module and the barrel, which can cause subtle initialization-order issues with bundlers/ts-node and is inconsistent with the rest of this folder where sibling files import each other directly (e.g. `from './rfq.js'`).

<details>

```suggestion
import type { RFQSettlementUnfilledActionType } from './rfq.js'
```

</details>

<details>

In `packages/sdk-ts/src/client/indexer/types/rfq-gw.ts` at line 2, `RFQSettlementUnfilledActionType` is imported from the barrel `'../types'`, which re-exports this very file and creates a circular import. Change the import to the concrete sibling path `'./rfq.js'` to match the convention used by other type files in this directory.

</details>

packages/sdk-ts/src/client/base/BaseGrpcConsumer.ts
```diff
@@ -220,12 +220,16 @@ export default class BaseGrpcConsumer {
...
+
+      return await call.response
+    }
+
     try {
-      return await this.retry(async () => {
-        const call = clientMethod(request, this.getRpcOptions())
-        return await call.response
-      })
+      return options?.noRetry ? await doCall() : await this.retry(doCall)
```
ivan-angjelkoski: 🔎 **Verification needed — Document `noRetry` semantics & confirm `retry()` has no auth/refresh side effects**

The new `options?: { noRetry?: boolean }` parameter defaults to retrying (matches the previous behavior — no regression for existing callers). However:

- The new `IndexerGrpcRfqGwApi.fetchPrepareAutoSign` call site uses `{ noRetry: true }`. This is correct: replaying an auto-sign prep with the same nonce/sequence could produce duplicate signed txs, so disabling retry is the right call. But the option is now part of the protected base API and any subclass author needs to know **when** to opt out. Add a JSDoc spelling out the contract.
- Please also confirm `this.retry(...)` does not also handle anything beyond exponential backoff — e.g. an auth-token refresh on `UNAUTHENTICATED`, metadata re-injection, or transport reconnect. If `retry` is the only place those happen, `noRetry: true` would silently bypass them and the first-call failure would surface a transport error to the caller without recovery. If `retry` is purely backoff, this is fine as-is.

<details>

```suggestion
    /**
     * @param options.noRetry  If true, skips the default retry wrapper.
     *   Pass `true` for non-idempotent RPCs where a server-side success
     *   followed by a transport error must not be retried (e.g. tx broadcast,
     *   prepare-auto-sign).
     */
```

</details>

<details>

In `packages/sdk-ts/src/client/base/BaseGrpcConsumer.ts` around lines 223 to 232, document the new `noRetry` option with a JSDoc explaining it should be set to `true` for non-idempotent RPCs (broadcast, prepare-auto-sign, etc.) where retrying after a server-side success could cause duplicate side effects. Additionally verify that `this.retry(doCall)` only performs retry/backoff and does not also handle auth-token refresh, metadata re-injection, or transport reconnect; if it does, refactor so the `noRetry` path still benefits from those (e.g. wrap `doCall` in the same auth/refresh logic and only skip the backoff loop).

</details>

PR #651, opened 2026-04-27
chore: keplr evm

ivan-angjelkoski:
This pull request introduces support for the Keplr EVM wallet across the codebase, improves EVM network switching (including auto-adding unknown networks), and makes several related type and error handling improvements. The main changes are grouped below.

**Keplr EVM Wallet Integration:**

- Added support for the Keplr EVM wallet in the wallet strategy factory, allowing it to be used as an EVM wallet (`Wallet.KeplrEvm`) (`packages/wallets/wallet-strategy/src/strategy/index.ts`).
- Implemented detection and retrieval of the Keplr EVM provider from the browser window, updating type definitions and provider utility functions to support `window.keplr.ethereum` (`packages/wallets/wallet-base/src/types/provider.ts`, `packages/wallets/wallet-evm/src/strategy/utils/keplrEvm.ts`, `packages/wallets/wallet-evm/src/utils/index.ts`) [[1]](diffhunk://#diff-ec5f32bc0e63a1bd526253c0343149b19e3a38410795dba08e56a56d650edf34L16-R20) [[2]](diffhunk://#diff-ec5f32bc0e63a1bd526253c0343149b19e3a38410795dba08e56a56d650edf34R29) [[3]](diffhunk://#diff-523e12c70052fc77813420bf26cded7338b016ad95ddecf93fc7a6c51f13bbc0L47-R56) [[4]](diffhunk://#diff-25d1cef4a02249528340458942d97d7f7bf701010ce30bf107d0648273ffce61L3-R12) [[5]](diffhunk://#diff-25d1cef4a02249528340458942d97d7f7bf701010ce30bf107d0648273ffce61R54-R57).

**EVM Network Switching Improvements:**

- Enhanced the `updateEvmNetwork` function to automatically add a new EVM network to the wallet if it is not already present, using chain configuration from `getEvmChainConfig` (`packages/wallets/wallet-evm/src/utils/index.ts`).
- Added explicit configuration for the Injective Testnet EVM chain in `getEvmChainConfig` (`packages/wallets/wallet-base/src/utils/viem.ts`).

**Error Handling and Naming Consistency:**

- Updated the Keplr EVM exception class to use a more accurate class name and simplified error string cleanup logic (`packages/exceptions/src/exceptions/exceptions/KeplrEvmException.ts`).
- Generalized error handling in EVM wallet strategy to use the correct exception class (`packages/wallets/wallet-evm/src/strategy/strategy.ts`).

## Summary by CodeRabbit

* **New Features**
  * Keplr EVM wallet support (now integrated into the EVM wallet flow).
  * Explicit testnet EVM chain handling so testnet networks are recognized.

* **Bug Fixes**
  * Improved network switching: 30s timeout, event-aware flow, and automatic chain addition when needed.
  * Better signing error routing for clearer failures.
  * More reliable provider detection and simplified Keplr error messages.

### DIFF
--- a/packages/exceptions/src/exceptions/exceptions/KeplrEvmException.ts
+++ b/packages/exceptions/src/exceptions/exceptions/KeplrEvmException.ts
@@ -4,13 +4,10 @@ import { mapErrorMessage } from '../utils/maps.js'
 import type { ErrorContext } from '../types/index.js'
 
 const removeKeplrFromErrorString = (message: string): string =>
-  message
-    .replaceAll('Keplr', '')
-    .replaceAll('Keplr', '')
-    .replaceAll('Keplr:', '')
+  message.replaceAll('Keplr:', '').replaceAll('Keplr', '')
 
 export class KeplrEvmException extends ConcreteException {
-  public static errorClass: string = 'KeplrException'
+  public static errorClass: string = 'KeplrEvmException'
 
   constructor(error: Error, context?: ErrorContext) {
     super(error, context)
--- a/packages/wallets/wallet-evm/src/strategy/utils/keplrEvm.ts
+++ b/packages/wallets/wallet-evm/src/strategy/utils/keplrEvm.ts
@@ -44,7 +44,7 @@ function getKeplrEvmFromWindow() {
   const injectedProviderExist =
     typeof window !== 'undefined' &&
     (typeof $window.ethereum !== 'undefined' ||
-      typeof ($window as any).keplr?.ethereum !== 'undefined')
+      typeof $window.keplr?.ethereum !== 'undefined')
 
   // No injected providers exist.
   if (!injectedProviderExist) {
@@ -52,8 +52,8 @@ function getKeplrEvmFromWindow() {
   }
 
   // Keplr exposes its EVM provider via window.keplr.ethereum
-  if (($window as any).keplr?.ethereum) {
-    return ($window as any).keplr.ethereum as BrowserEip1993Provider
+  if ($window.keplr?.ethereum) {
+    return $window.keplr.ethereum
   }
 
   if ($window.ethereum?.isKeplr) {
--- a/packages/wallets/wallet-evm/src/strategy/utils/keplrEvm.ts
+++ b/packages/wallets/wallet-evm/src/strategy/utils/keplrEvm.ts
@@ -44,7 +44,7 @@ function getKeplrEvmFromWindow() {
   const injectedProviderExist =
     typeof window !== 'undefined' &&
     (typeof $window.ethereum !== 'undefined' ||
-      typeof $window.keplr?.ethereum !== 'undefined')
+      typeof ($window as any).keplr?.ethereum !== 'undefined')
 
   // No injected providers exist.
   if (!injectedProviderExist) {
@@ -52,8 +52,8 @@ function getKeplrEvmFromWindow() {
   }
 
   // Keplr exposes its EVM provider via window.keplr.ethereum
-  if ($window.keplr?.ethereum) {
-    return $window.keplr.ethereum
+  if (($window as any).keplr?.ethereum) {
+    return ($window as any).keplr.ethereum as BrowserEip1993Provider
   }
 
   if ($window.ethereum?.isKeplr) {
--- a/packages/wallets/wallet-evm/src/utils/index.ts
+++ b/packages/wallets/wallet-evm/src/utils/index.ts
@@ -1,10 +1,15 @@
 import { capitalize } from '@injectivelabs/utils'
 import { WalletException } from '@injectivelabs/exceptions'
-import { Wallet, isEvmBrowserWallet } from '@injectivelabs/wallet-base'
+import {
+  Wallet,
+  getEvmChainConfig,
+  isEvmBrowserWallet,
+} from '@injectivelabs/wallet-base'
 import { getRabbyProvider } from '../strategy/utils/rabby.js'
 import { getOkxWalletProvider } from '../strategy/utils/Okx.js'
 import { getBitGetProvider } from '../strategy/utils/bitget.js'
 import { getPhantomProvider } from '../strategy/utils/phantom.js'
+import { getKeplrEvmProvider } from '../strategy/utils/keplrEvm.js'
 import { getMetamaskProvider } from '../strategy/utils/metamask.js'
 import { getTrustWalletProvider } from '../strategy/utils/trustWallet.js'
 import type { EvmChainId } from '@injectivelabs/ts-types'
@@ -46,6 +51,10 @@ export const getEvmProvider = async (
       provider = (await getOkxWalletProvider()) as BrowserEip1993Provider
     }
 
+    if (wallet === Wallet.KeplrEvm) {
+      provider = (await getKeplrEvmProvider()) as BrowserEip1993Provider
+    }
+
     if (!provider) {
       throw new WalletException(
         new Error(`Please install ${capitalize(wallet)} Extension`),
@@ -75,23 +84,57 @@ export const updateEvmNetwork = async (wallet: Wallet, chainId: EvmChainId) => {
     )
   }
 
-  try {
-    const chainIdToHex = chainId.toString(16)
+  const chainIdToHex = `0x${chainId.toString(16)}`
 
+  try {
     return await Promise.race([
       provider.request({
         method: 'wallet_switchEthereumChain',
-        params: [{ chainId: `0x${chainIdToHex}` }],
+        params: [{ chainId: chainIdToHex }],
       }),
       new Promise<void>((resolve) =>
         provider.on('chainChanged', ({ chain }: any) => {
-          if (chain?.id === chainIdToHex) {
+          if (chain?.id === chainId.toString(16)) {
             resolve()
           }
         }),
       ),
     ])
-  } catch {
+  } catch (switchError: any) {
+    const errorCode =
+      switchError?.code ?? switchError?.data?.originalError?.code
+
+    if (errorCode === 4902) {
+      const chainConfig = getEvmChainConfig(chainId)
+
+      try {
+        const rpcUrl = chainConfig.rpcUrls?.default?.http?.[0]
+        const explorerUrl =
+          chainConfig.blockExplorers?.default?.url || undefined
+
+        await provider.request({
+          method: 'wallet_addEthereumChain',
+          params: [
+            {
+              chainId: chainIdToHex,
+              chainName: chainConfig.name,
+              nativeCurrency: chainConfig.nativeCurrency,
+              rpcUrls: rpcUrl ? [rpcUrl] : [],
+              blockExplorerUrls: explorerUrl ? [explorerUrl] : undefined,
+            },
+          ],
+        })
+
+        return
+      } catch {
+        throw new WalletException(
+          new Error(
+            `Failed to add ${chainConfig.name} network to ${capitalize(wallet)}`,
+          ),
+        )
+      }
+    }
+
     throw new WalletException(
       new Error(`Please update your ${capitalize(wallet)} network`),
     )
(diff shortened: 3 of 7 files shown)


packages/wallets/wallet-evm/src/utils/index.ts
```diff
@@ -75,23 +84,57 @@ export const updateEvmNetwork = async (wallet: Wallet, chainId: EvmChainId) => {
...
+        params: [{ chainId: chainIdToHex }],
       }),
       new Promise<void>((resolve) =>
         provider.on('chainChanged', ({ chain }: any) => {
-          if (chain?.id === chainIdToHex) {
+          if (chain?.id === chainId.toString(16)) {
             resolve()
           }
         }),
       ),
```
ivan-angjelkoski: ⚠️ **Potential issue — `chainChanged` listener never resolves and leaks on every call**

The `Promise.race` here registers a `chainChanged` listener that can't fire as written:

1. EIP-1193 emits `chainChanged` with the chain ID as a hex string directly (`"0x59f"`), not an object. Destructuring `({ chain }: any)` makes `chain` `undefined`, so `chain?.id === chainId.toString(16)` is always `false` and the inner `Promise` never resolves.
2. Even if the payload were an object, the comparison strips the `0x` prefix from one side (`chainId.toString(16)`) but the wallet emits it with the prefix.
3. The listener is added with `provider.on(...)` but never removed — every call to `updateEvmNetwork` adds a new listener that lives forever.

In practice the race resolves only via `wallet_switchEthereumChain`, but the leaked listener accumulates. Compare with the working pattern in `strategy.ts addEvmNetwork` (lines 467-485) which uses `(newChainId: string) => ...`, compares with `0x` prefix case-insensitively, and cleans up via `removeListener` + `clearTimeout`.

<details>

```suggestion
      new Promise<void>((resolve) => {
        const handleChainChanged = (newChainId: string) => {
          if (
            typeof newChainId === 'string' &&
            newChainId.toLowerCase() === chainIdToHex.toLowerCase()
          ) {
            provider.removeListener('chainChanged', handleChainChanged)
            resolve()
          }
        }

        provider.on('chainChanged', handleChainChanged)
      }),
```

</details>

<details>

```

In `@packages/wallets/wallet-evm/src/utils/index.ts` around lines 95 - 101,
the `chainChanged` listener inside the `Promise.race` is unreachable: it
destructures `({ chain }: any)` but EIP-1193 emits the chain ID as a hex
string directly, and it compares to `chainId.toString(16)` (no `0x`) while
wallets emit the prefixed form. The listener is also never removed, leaking
on every call. Replace the listener with `(newChainId: string) => ...`,
compare case-insensitively against `chainIdToHex`, and call
`provider.removeListener('chainChanged', handleChainChanged)` once it
resolves — mirror the cleanup pattern already used in
`strategy.ts addEvmNetwork` around lines 467 - 485.
```

</details>

packages/wallets/wallet-evm/src/utils/index.ts
```diff
@@ -75,23 +84,57 @@ export const updateEvmNetwork = async (wallet: Wallet, chainId: EvmChainId) => {
...
+        })
+
+        return
+      } catch {
+        throw new WalletException(
+          new Error(
+            `Failed to add ${chainConfig.name} network to ${capitalize(wallet)}`,
+          ),
+        )
+      }
```
ivan-angjelkoski, line 150: 🛠️ **Refactor suggestion — preserve the underlying error when `wallet_addEthereumChain` fails**

The bare `catch {` swallows the original error from `provider.request(...)`, so users (and Sentry) only see `Failed to add ${chainConfig.name} network to ${capitalize(wallet)}` with no clue why. Common failure modes (user rejected, malformed `nativeCurrency`, RPC unreachable, EIP-3085 validation error) all collapse to the same string. Capture the error and propagate its message so the surface error is debuggable.

<details>

```suggestion
      } catch (addError: any) {
        throw new WalletException(
          new Error(
            `Failed to add ${chainConfig.name} network to ${capitalize(
              wallet,
            )}: ${addError?.message ?? 'unknown error'}`,
          ),
        )
      }
```

</details>

<details>

```

In `@packages/wallets/wallet-evm/src/utils/index.ts` around lines 129 - 135,
the bare `catch {` discards the underlying error from
`wallet_addEthereumChain`, so all add-chain failures surface as the same
generic message. Bind the error (`catch (addError: any)`) and append
`addError?.message ?? 'unknown error'` to the WalletException message so
EIP-3085 validation errors, user rejection, and RPC failures stay
distinguishable in logs and UI.
```

</details>

ivan-angjelkoski (commented):
## Review summary

Four findings, focused on the new `updateEvmNetwork` flow plus one regression-watch:

- **Critical** — `updateEvmNetwork`'s `chainChanged` listener can never fire (wrong EIP-1193 payload shape, missing `0x` prefix on the comparison) and is never removed, leaking on every call.
- **Medium** — The nested `catch {` after `wallet_addEthereumChain` discards the underlying error, collapsing every add-chain failure (user rejection, malformed `nativeCurrency`, RPC unreachable) into the same generic message.
- **Medium** — `rpcUrls: rpcUrl ? [rpcUrl] : []` lets an empty array reach the wallet, which violates EIP-3085 and produces an opaque wallet error masked by the generic catch above.
- **Verification** — `KeplrEvmException.errorClass` is renamed from `'KeplrException'` to `'KeplrEvmException'`. Semantically correct, but the value flows into `Error#name`, so any consumer matching on `'KeplrException'` (UI banners, Sentry tags, telemetry) silently stops matching once this ships.

Security and performance lenses turned up no findings beyond the listener leak above.

**Positives:** the `removeKeplrFromErrorString` ordering fix (running `Keplr:` before `Keplr`) is a clean correctness improvement; the typed `WindowWithEip1193Provider.keplr` field replaces a stack of `($window as any)` casts in `keplrEvm.ts` and is the right shape; routing `EvmWallet.signEip712TypedData`'s catch through `this.EvmWalletException(...)` brings KeplrEvm in line with the other EVM wallets; and the `Wallet.KeplrEvm` strategy-factory case slots cleanly into the existing EVM switch.

packages/wallets/wallet-evm/src/utils/index.ts
```diff
@@ -75,23 +84,57 @@ export const updateEvmNetwork = async (wallet: Wallet, chainId: EvmChainId) => {
...
+          chainConfig.blockExplorers?.default?.url || undefined
+
+        await provider.request({
+          method: 'wallet_addEthereumChain',
+          params: [
+            {
+              chainId: chainIdToHex,
+              chainName: chainConfig.name,
+              nativeCurrency: chainConfig.nativeCurrency,
+              rpcUrls: rpcUrl ? [rpcUrl] : [],
```
ivan-angjelkoski, line 137: 🛠️ **Refactor suggestion — empty `rpcUrls` array violates EIP-3085 and surfaces as a confusing wallet error**

`rpcUrls: rpcUrl ? [rpcUrl] : []` defends against `chainConfig.rpcUrls?.default?.http?.[0]` being undefined, but EIP-3085 (and every wallet implementation that follows it) requires `rpcUrls` to be a **non-empty** array. If the lookup falls through, the wallet rejects the request with an opaque message that surfaces via the catch block above as `Failed to add … network to …`, masking the real cause.

Better to fail fast with a clear message before calling the provider — that way the caller and any error tracking pipeline get a precise diagnosis rather than a generic add-chain failure.

<details>

```suggestion
      try {
        const rpcUrl = chainConfig.rpcUrls?.default?.http?.[0]
        const explorerUrl =
          chainConfig.blockExplorers?.default?.url || undefined

        if (!rpcUrl) {
          throw new WalletException(
            new Error(
              `Cannot add ${chainConfig.name} to ${capitalize(
                wallet,
              )}: no RPC URL configured for chainId ${chainId}`,
            ),
          )
        }

        await provider.request({
          method: 'wallet_addEthereumChain',
          params: [
            {
              chainId: chainIdToHex,
              chainName: chainConfig.name,
              nativeCurrency: chainConfig.nativeCurrency,
              rpcUrls: [rpcUrl],
              blockExplorerUrls: explorerUrl ? [explorerUrl] : undefined,
            },
          ],
        })

        return
```

</details>

<details>

```

In `@packages/wallets/wallet-evm/src/utils/index.ts` around lines 110 - 128,
the `wallet_addEthereumChain` payload passes `rpcUrls: rpcUrl ? [rpcUrl] : []`
which lets an empty array reach the wallet — EIP-3085 requires a non-empty
array, and wallets reject the call with an opaque error that the surrounding
catch then re-wraps as the generic 'Failed to add … network'. Add an
explicit guard before the request: when `rpcUrl` is falsy throw a
`WalletException` with a message that names the chain and chainId so the
real cause is visible. Then pass `rpcUrls: [rpcUrl]` unconditionally.
```

</details>

packages/exceptions/src/exceptions/exceptions/KeplrEvmException.ts
```diff
@@ -4,13 +4,10 @@ import { mapErrorMessage } from '../utils/maps.js'
...
 const removeKeplrFromErrorString = (message: string): string =>
-  message
-    .replaceAll('Keplr', '')
-    .replaceAll('Keplr', '')
-    .replaceAll('Keplr:', '')
+  message.replaceAll('Keplr:', '').replaceAll('Keplr', '')
 
 export class KeplrEvmException extends ConcreteException {
-  public static errorClass: string = 'KeplrException'
+  public static errorClass: string = 'KeplrEvmException'
```
ivan-angjelkoski, line 10: 🔎 **Verification needed — `KeplrEvmException.errorClass` rename is observable to consumers**

`errorClass` flips from `'KeplrException'` to `'KeplrEvmException'`. The class itself is `KeplrEvmException`, so the new value is semantically correct, but `parse()` calls `this.setName(KeplrEvmException.errorClass)` so the rename also changes the resulting `Error#name`. Any downstream code that switches on `error.name === 'KeplrException'` (UI error banners, Sentry tag rules, telemetry filters) will silently stop matching once this ships.

Worth confirming:
1. No consumer code (Helix, true-current, hub, bridge, fe-server) keys off `'KeplrException'` for KeplrEvm errors.
2. If this is the intended fix for the prior misnaming, callers should be updated together — and the rename should be flagged in the changelog so internal teams aren't surprised.

No committable change here — just confirm the blast radius before merging.

<details>

```

In `@packages/exceptions/src/exceptions/exceptions/KeplrEvmException.ts`
around line 10, `errorClass` was renamed from 'KeplrException' to
'KeplrEvmException', which is propagated to `Error#name` via `setName` in
`parse()`. Search the consuming repos (injective-helix, injective-true-current,
injective-hub, injective-bridge, injective-fe-server) for string matches on
'KeplrException' and confirm none of them branch on the old name; if any
do, update them in lockstep. Add a note to the changelog so app teams
relying on `error.name` are aware the wire value changed.
```

</details>

packages/wallets/wallet-evm/src/strategy/utils/keplrEvm.ts
```diff
@@ -44,16 +44,16 @@ function getKeplrEvmFromWindow() {
...
   // No injected providers exist.
   if (!injectedProviderExist) {
     return
   }
 
   // Keplr exposes its EVM provider via window.keplr.ethereum
-  if (($window as any).keplr?.ethereum) {
-    return ($window as any).keplr.ethereum as BrowserEip1993Provider
+  if ($window.keplr?.ethereum) {
+    return $window.keplr.ethereum
```
ThomasRalee: Lets do this for consistency:
```
return ($window as any).keplr.ethereum as BrowserEip1993Provider
```

packages/wallets/wallet-evm/src/strategy/utils/keplrEvm.ts
```diff
@@ -44,16 +44,16 @@ function getKeplrEvmFromWindow() {
   const injectedProviderExist =
     typeof window !== 'undefined' &&
     (typeof $window.ethereum !== 'undefined' ||
-      typeof ($window as any).keplr?.ethereum !== 'undefined')
+      typeof $window.keplr?.ethereum !== 'undefined')
```
ThomasRalee: Let's use this for consistency:

` typeof ($window as any).keplr?.ethereum !== 'undefined') `

PR #663, opened 2026-05-05
feat: exchange v2 messages

ThomasRalee:
## Summary by CodeRabbit

* **New Features**
  * SDK now exposes a broad set of Exchange V2 messages (deposits, withdrawals, spot/derivative/binary orders, batch ops, margin adjustments, market launches, liquidations, stake grants, reclaim/opt-out, sign/external transfers) with full serialization and signing adapters.

* **EIP‑712 / Web3 Gateway**
  * Added EIP‑712 v2 compatibility and improved Web3 Gateway mappings (including expiration_block handling).

* **Tests**
  * Comprehensive unit and EIP‑712 integration tests validating new V2 message formats and signing paths.

### DIFF
--- a/packages/sdk-ts/src/core/modules/exchange/msgs/MsgAdminUpdateBinaryOptionsMarketV2.ts
+++ b/packages/sdk-ts/src/core/modules/exchange/msgs/MsgAdminUpdateBinaryOptionsMarketV2.ts
@@ -1,4 +1,5 @@
 import { toChainFormat } from '@injectivelabs/utils'
+import { GeneralException } from '@injectivelabs/exceptions'
 import * as InjectiveExchangeV2TxPb from '@injectivelabs/core-proto-ts-v2/generated/injective/exchange/v2/tx_pb'
 import * as InjectiveExchangeV2MarketPb from '@injectivelabs/core-proto-ts-v2/generated/injective/exchange/v2/market_pb'
 import { MsgBase } from '../../MsgBase.js'
@@ -91,19 +92,12 @@ export default class MsgAdminUpdateBinaryOptionsMarketV2 extends MsgBase<
     }
   }
 
-  public toEip712() {
-    const amino = this.toAmino()
-    const { type, value } = amino
-
-    const messageAdjusted = {
-      ...value,
-      settlement_price: toChainFormat(value.settlement_price).toFixed(),
-    }
-
-    return {
-      type,
-      value: messageAdjusted,
-    }
+  public toEip712(): never {
+    throw new GeneralException(
+      new Error(
+        'EIP712 v1 is not supported for MsgAdminUpdateBinaryOptionsMarket, use toEip712V2 instead.',
+      ),
+    )
   }
 
   public toEip712V2() {
--- a/packages/sdk-ts/src/core/modules/exchange/msgs/MsgAdminUpdateBinaryOptionsMarketV2.ts
+++ b/packages/sdk-ts/src/core/modules/exchange/msgs/MsgAdminUpdateBinaryOptionsMarketV2.ts
@@ -1,0 +1,136 @@
+import { toChainFormat } from '@injectivelabs/utils'
+import * as InjectiveExchangeV2TxPb from '@injectivelabs/core-proto-ts-v2/generated/injective/exchange/v2/tx_pb'
+import * as InjectiveExchangeV2MarketPb from '@injectivelabs/core-proto-ts-v2/generated/injective/exchange/v2/market_pb'
+import { MsgBase } from '../../MsgBase.js'
+import { numberToCosmosSdkDecString } from '../../../../utils/numbers.js'
+
+export declare namespace MsgAdminUpdateBinaryOptionsMarketV2 {
+  export interface Params {
+    sender: string
+    marketId: string
+    settlementPrice: string
+    expirationTimestamp: string
+    settlementTimestamp: string
+    status: InjectiveExchangeV2MarketPb.MarketStatus
+  }
+
+  export type Proto = InjectiveExchangeV2TxPb.MsgAdminUpdateBinaryOptionsMarket
+}
+
+const createMessage = (params: MsgAdminUpdateBinaryOptionsMarketV2.Params) => {
+  const message =
+    InjectiveExchangeV2TxPb.MsgAdminUpdateBinaryOptionsMarket.create({
+      sender: params.sender,
+      marketId: params.marketId,
+      settlementPrice: params.settlementPrice,
+      expirationTimestamp: BigInt(params.expirationTimestamp),
+      settlementTimestamp: BigInt(params.settlementTimestamp),
+      status: params.status,
+    })
+
+  return message
+}
+
+/**
+ * @category Messages
+ */
+export default class MsgAdminUpdateBinaryOptionsMarketV2 extends MsgBase<
+  MsgAdminUpdateBinaryOptionsMarketV2.Params,
+  MsgAdminUpdateBinaryOptionsMarketV2.Proto
+> {
+  static fromJSON(
+    params: MsgAdminUpdateBinaryOptionsMarketV2.Params,
+  ): MsgAdminUpdateBinaryOptionsMarketV2 {
+    return new MsgAdminUpdateBinaryOptionsMarketV2(params)
+  }
+
+  public toProto() {
+    const { params: initialParams } = this
+
+    const params = {
+      ...initialParams,
+      settlementPrice: toChainFormat(initialParams.settlementPrice).toFixed(),
+    } as MsgAdminUpdateBinaryOptionsMarketV2.Params
+
+    return createMessage(params)
+  }
+
+  public toData() {
+    const proto = this.toProto()
+
+    return {
+      '@type': '/injective.exchange.v2.MsgAdminUpdateBinaryOptionsMarket',
+      ...proto,
+    }
+  }
+
+  public toAmino() {
+    const { params } = this
+    const message = {
+      sender: params.sender,
+      market_id: params.marketId,
+      settlement_price: params.settlementPrice,
+      expiration_timestamp: params.expirationTimestamp,
+      settlement_timestamp: params.settlementTimestamp,
+      status: params.status,
+    }
+
+    return {
+      type: 'exchange/MsgAdminUpdateBinaryOptionsMarket',
+      value: message,
+    }
+  }
+
+  public toWeb3Gw() {
+    const amino = this.toAmino()
+    const { value } = amino
+
+    return {
+      '@type': '/injective.exchange.v2.MsgAdminUpdateBinaryOptionsMarket',
+      ...value,
+    }
+  }
+
+  public toEip712() {
+    const amino = this.toAmino()
+    const { type, value } = amino
+
+    const messageAdjusted = {
+      ...value,
+      settlement_price: toChainFormat(value.settlement_price).toFixed(),
+    }
+
+    return {
+      type,
+      value: messageAdjusted,
+    }
+  }
+
+  public toEip712V2() {
+    const { params } = this
+    const web3gw = this.toWeb3Gw()
+
+    const messageAdjusted = {
+      ...web3gw,
+      settlement_price: numberToCosmosSdkDecString(params.settlementPrice),
+      status: InjectiveExchangeV2MarketPb.MarketStatus[params.status],
+    }
+
+    return messageAdjusted
+  }
+
+  public toDirectSign() {
+    const proto = this.toProto()
+
+    return {
+      type: '/injective.exchange.v2.MsgAdminUpdateBinaryOptionsMarket',
+      message: proto,
+    }
+  }
+
+  public toBinary(): Uint8Array {
+    return InjectiveExchangeV2TxPb.MsgAdminUpdateBinaryOptionsMarket.toBinary(
+      this.toProto(),
+    )
+  }
+}
--- a/packages/sdk-ts/src/core/modules/exchange/msgs/MsgBatchUpdateOrdersV2.ts
+++ b/packages/sdk-ts/src/core/modules/exchange/msgs/MsgBatchUpdateOrdersV2.ts
@@ -13,6 +13,7 @@ interface SpotOrderToCreate {
   price: string
   quantity: string
   cid?: string
+  expirationBlock?: string
 }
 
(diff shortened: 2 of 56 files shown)


packages/sdk-ts/src/core/modules/exchange/msgs/MsgInstantBinaryOptionsMarketLaunchV2.ts
```diff
@@ -0,0 +1,199 @@
...
+      ),
+      min_quantity_tick_size: numberToCosmosSdkDecString(
+        params.market.minQuantityTickSize,
+      ),
+      min_notional: numberToCosmosSdkDecString(params.market.minNotional),
+      taker_fee_rate: numberToCosmosSdkDecString(params.market.takerFeeRate),
+      maker_fee_rate: numberToCosmosSdkDecString(params.market.makerFeeRate),
+      oracle_type:
+        InjectiveOracleV1Beta1OraclePb.OracleType[params.market.oracleType],
+      open_notional_cap: {},
```
ivan-angjelkoski, line 167: ⚠️ **Potential issue — `MsgInstantBinaryOptionsMarketLaunchV2` silently drops the new `open_notional_cap` proto field**

The v2 proto `injective.exchange.v2.MsgInstantBinaryOptionsMarketLaunch` adds a new field at position 16:

```ts
openNotionalCap?: OpenNotionalCap;
```

(see `core-proto-ts-v2/generated/injective/exchange/v2/tx_pb.d.ts`). The v1beta1 message did not have this field — it is one of the headline additions in v2.

In this V2 SDK message the field is missing from `Params`, never assigned in `toProto()`, never emitted in `toAmino()`/`toWeb3Gw()`, and only appears in `toEip712V2()` as a hard-coded empty placeholder `open_notional_cap: {}`. Net effect:

1. Callers cannot pass an `openNotionalCap` value; the proto field is permanently `undefined`.
2. `toEip712V2()` and `toWeb3Gw()` disagree on shape — the v2 typed-data schema declares an `open_notional_cap` field that the broadcast amino value never includes, so signed bytes won't match the tx that actually goes on-chain when a non-empty cap is set.
3. Any future caller that expects the SDK to round-trip a non-empty cap will silently lose data.

Either expose `openNotionalCap` in `Params` and wire it through `toProto`/`toAmino`/`toWeb3Gw`/`toEip712V2`, or — if the chain currently mandates an empty cap — leave a comment explaining the placeholder and add a `verifyOpenNotionalCapNotSet` style check so we don't accumulate this drift across V2 launches (`MsgUpdateDerivativeMarketV2` already uses the same `new_open_notional_cap: {}` placeholder, so the precedent is mixed).

<details>

```

In `@packages/sdk-ts/src/core/modules/exchange/msgs/MsgInstantBinaryOptionsMarketLaunchV2.ts`
around lines 7 - 199, the new v2 proto field
`MsgInstantBinaryOptionsMarketLaunch.openNotionalCap` (field 16) is dropped:
the `Params` interface does not expose it, `createMessage`/`toProto` never
sets it, `toAmino`/`toWeb3Gw` omit it, and `toEip712V2` only inserts a
hard-coded `open_notional_cap: {}`. Add an optional `openNotionalCap`
field to `Params` (importing `OpenNotionalCap` from
`@injectivelabs/core-proto-ts-v2/generated/injective/exchange/v2/tx_pb`),
forward it through `createMessage` (`openNotionalCap: params.openNotionalCap`),
include `open_notional_cap: params.openNotionalCap` in `toAmino`/`toWeb3Gw`,
and replace the empty placeholder in `toEip712V2` with the user-supplied
value (or `{}` only when `params.openNotionalCap` is undefined). Update
`MsgInstantBinaryOptionsMarketLaunchV2.spec.ts` to assert the field
round-trips. Confirm with the chain team whether
`MsgUpdateDerivativeMarketV2`'s identical `new_open_notional_cap: {}`
placeholder should be addressed in the same way.
```

</details>

ivan-angjelkoski (commented):
## Review summary

Three findings against the new V2 exchange message bundle:

- **Critical** — `MsgInstantBinaryOptionsMarketLaunchV2` drops the new `open_notional_cap` proto field (field 16); `Params` doesn't expose it, the proto/amino values omit it, and only `toEip712V2()` carries an empty `{}` placeholder. Callers cannot set a non-empty cap, and the EIP712 schema diverges from the broadcast value when one *is* set.
- **Verification** — `expirationBlock` is hard-coded to `0` across every V2 create-order code path (`MsgBatchUpdateOrdersV2` + the six `MsgCreate*V2` files + `MsgLiquidatePositionV2`). The proto field exists; please confirm whether the v2 chain handler currently honours it. If yes, expose it in `Params`; if no, leave a one-line comment so the gap is intentional and discoverable.
- **Refactor** — `MsgAdminUpdateBinaryOptionsMarketV2.toEip712()` still returns typed data the chain has already rejected (`spec.ts` skips the V1 round-trip with a `*math.LegacyDec` comment). Mirror `MsgReclaimLockedFundsV2` and throw a `GeneralException` with a clear "use V2" message instead of returning broken bytes.

**Positives:** the per-message spec coverage is thorough — every new file has matching V1 + V2 EIP712 round-trip tests against the chain endpoint, the `'/injective.exchange.v2.*'` type-URL strings are consistent across `toData`/`toAmino`/`toEip712V2`/`toDirectSign`, and the `maps.ts` additions cover all 26 new amino mappings plus the `expiration_block` reflection rule. The pattern carry-over from V1 is clean for the simple messages (`MsgDepositV2`, `MsgWithdrawV2`, `MsgCancelSpotOrderV2`, etc.) and the `MsgBatchUpdateOrdersV2.toEip712Types()` schema-augment workaround is a reasonable solution to the chain's expiration-block schema quirk.

packages/sdk-ts/src/core/modules/exchange/msgs/MsgBatchUpdateOrdersV2.ts
```diff
@@ -0,0 +1,872 @@
...
+    quantity: args.quantity,
+    cid: args.cid,
+  })
+
+  const order = InjectiveExchangeV2OrderPb.SpotOrder.create({
+    marketId: args.marketId,
+    orderType: args.orderType,
+    orderInfo: orderInfo,
+    triggerPrice: args.triggerPrice,
+    expirationBlock: BigInt(0),
```
ivan-angjelkoski: 🔎 **Verification needed — `expirationBlock` is hard-coded to `0` across every V2 create-order path**

The v2 proto `SpotOrder` and `DerivativeOrder` types both expose `expirationBlock: bigint` (a real chain-supported field). This PR introduces eight V2 messages that build orders, and every one of them silently pins it to zero:

- `MsgBatchUpdateOrdersV2.ts`:92, 115, 138 (`createSpotOrder`, `createDerivativeOrder`, `createBinaryOptionOrder`)
- `MsgCreateSpotLimitOrderV2.ts`:37
- `MsgCreateSpotMarketOrderV2.ts` (same pattern)
- `MsgCreateDerivativeLimitOrderV2.ts` (same pattern)
- `MsgCreateDerivativeMarketOrderV2.ts`:39
- `MsgCreateBinaryOptionsLimitOrderV2.ts` (same pattern)
- `MsgCreateBinaryOptionsMarketOrderV2.ts` (same pattern)
- `MsgLiquidatePositionV2.ts`:47-53 (the optional `order` field never receives `expirationBlock`)

Because none of the `Params` interfaces expose the field, callers using the V2 SDK can never produce an order with a non-zero expiration block, even though the chain proto carries the field. The matching `expiration_block: '0'` strings in `toAmino` / `toEip712V2` reinforce the lockout.

Two paths from here:

1. **If v2 chain logic does not yet honour `expiration_block`** — leave the hard-coding in place but add a one-line comment near each `expirationBlock: BigInt(0)` so the gap is intentional and discoverable; otherwise the next person to look will silently re-introduce it.
2. **If the chain does honour it** — expose `expirationBlock?: string` on every order-shaped `Params`, plumb it through `toProto`/`toAmino`/`toEip712V2`, and add at least one spec that round-trips a non-zero value.

Worth confirming with whoever wrote the v2 chain handler — this is the kind of feature gap that's invisible until a market maker asks "why doesn't the SDK support timed orders?".

<details>

```

In `@packages/sdk-ts/src/core/modules/exchange/msgs/MsgBatchUpdateOrdersV2.ts`
around lines 76 - 142, the helpers `createSpotOrder`, `createDerivativeOrder`,
and `createBinaryOptionOrder` hard-code `expirationBlock: BigInt(0)` and the
matching `toAmino`/`toEip712V2` paths emit `expiration_block: '0'`. The same
pattern exists in `MsgCreateSpotLimitOrderV2.ts`,
`MsgCreateSpotMarketOrderV2.ts`, `MsgCreateDerivativeLimitOrderV2.ts`,
`MsgCreateDerivativeMarketOrderV2.ts`, `MsgCreateBinaryOptionsLimitOrderV2.ts`,
`MsgCreateBinaryOptionsMarketOrderV2.ts`, and inside
`MsgLiquidatePositionV2.ts` `createMessage`. Confirm with the chain team
whether `expiration_block` is currently honoured for v2; if it is, add an
optional `expirationBlock?: string` field to each order-shaped `Params`,
thread it through (`expirationBlock: BigInt(params.expirationBlock || 0)`
in proto helpers and `expiration_block: params.expirationBlock || '0'` in
amino/eip712), and add a spec that round-trips a non-zero value. If it is
not honoured yet, add a short comment next to each `BigInt(0)` /
`expiration_block: '0'` site explaining the chain limitation so the next
reader doesn't waste time re-discovering it.
```

</details>

packages/sdk-ts/src/core/modules/exchange/msgs/MsgAdminUpdateBinaryOptionsMarketV2.ts
```diff
@@ -0,0 +1,136 @@
...
+    const messageAdjusted = {
+      ...value,
+      settlement_price: toChainFormat(value.settlement_price).toFixed(),
+    }
+
+    return {
+      type,
+      value: messageAdjusted,
+    }
+  }
```
ivan-angjelkoski: 🛠️ **Refactor suggestion — `toEip712()` produces broken typed data the chain has already rejected; align it with `MsgReclaimLockedFundsV2` and throw**

`MsgAdminUpdateBinaryOptionsMarketV2.spec.ts` skips its own `EIP712 v1` round-trip with the explicit comment:

```ts
// not supported on chain (invalid Go type *math.LegacyDec for settlement_price)
it.skip('EIP712 v1', async () => {
```

…meaning the indexer's `prepareEip712Request` blows up on this message in V1 mode. But the class still publishes a public `toEip712()` (lines 257-270) that builds a `settlement_price: toChainFormat(...).toFixed()` value. A caller picking V1 EIP712 — directly or via a strategy that defaults to V1 — gets bytes that no chain endpoint can verify, with no signal that the path is unsupported.

`MsgReclaimLockedFundsV2` already establishes the precedent for this scenario:

```ts
public toEip712(): never {
  throw new GeneralException(
    new Error('EIP712 is not supported for MsgReclaimLockedFunds.'),
  )
}
```

Mirror that: replace the `toEip712()` body (and `toWeb3Gw` if it shares the same broken `settlement_price` shape — please verify) with a `throw new GeneralException(...)` calling out V1 as unsupported and pointing at `toEip712V2()`. That way the failure surfaces at the call site instead of as an opaque `prepareEip712Request` 4xx in production.

<details>

```suggestion
  public toEip712(): never {
    throw new GeneralException(
      new Error(
        'EIP712 v1 is not supported for MsgAdminUpdateBinaryOptionsMarket on chain (invalid Go type *math.LegacyDec for settlement_price). Use toEip712V2() instead.',
      ),
    )
  }
```

</details>

<details>

```

In `@packages/sdk-ts/src/core/modules/exchange/msgs/MsgAdminUpdateBinaryOptionsMarketV2.ts`
around lines 257 - 270, the `toEip712()` method silently produces typed
data that the chain already rejects (the spec file skips the V1 round-trip
with the comment "not supported on chain (invalid Go type *math.LegacyDec
for settlement_price)"). Mirror the pattern in
`MsgReclaimLockedFundsV2.ts` (lines 74 - 78): change the return type to
`never` and throw a `GeneralException` whose message names V1 as
unsupported and points callers at `toEip712V2()`. Add the
`import { GeneralException } from '@injectivelabs/exceptions'` line at the
top of the file. Also verify `toWeb3Gw()`/`toAmino()` for the same
`settlement_price` shape — if they are equally unsigned-on-chain in V1
mode, throw from those too.
```

</details>

ivan-angjelkoski (commented):
Additional inline findings (continuation of the review above).

ivan-angjelkoski (commented):
summary-only test

ivan-angjelkoski (commented):
## Review summary

Follow-up pass on the latest `fix: ai review` commit — four findings, one critical:

- **Critical** — `MsgInstantBinaryOptionsMarketLaunchV2.toEip712()` is still implemented even though the spec marks it as unsupported on chain (same root cause as `MsgAdminUpdateBinaryOptionsMarketV2`, which the fix commit correctly aligned to throw). It should mirror the throw.
- **Medium** — `MsgCancelSpotOrderV2.toProto()` throws raw `Error` for the new orderHash/cid guard; other V2 messages use `GeneralException` from `@injectivelabs/exceptions`.
- **Medium** — `expiration_block` types-schema patch is duplicated between `MsgBatchUpdateOrdersV2.toEip712Types()` and `MsgLiquidatePositionV2.toEip712Types()`. Extract to a shared helper before the third copy lands.
- **Verification** — `MsgBatchUpdateOrdersV2.toEip712V2()` always emits empty arrays for absent cancel/create lists. Spec only validates the populated-`spotOrdersToCreate` case against the chain; the empty-shape assumption is unverified for the other combinations.

**Positives** — the recent `fix: ai review` commit cleanly addressed the previous round (expiration_block plumbing, the `MsgAdmin*V2.toEip712` throw, removal of the `open_notional_cap: {}` placeholder, and the new orderHash/cid guard on `MsgCancelSpotOrderV2`). The buildSpotOrders/buildDerivativeOrders/buildBinaryOptionOrders helpers in `MsgBatchUpdateOrdersV2` are a nice tightening over the inline V1 pattern, and the `expirationBlock` round-trip spec on `MsgCreateDerivativeLimitOrderV2` is exactly the kind of regression guard worth keeping.

PR #702, opened 2026-07-13
fix: destroyed stream manager can be retried

ThomasRalee:
## Summary by CodeRabbit

- **New Features**
  - Added a terminal **Destroyed** stream state.
- **Bug Fixes**
  - Streams can no longer start/connect/retry after being destroyed, including during in-flight reconnect handling.
  - Destruction now performs immediate cleanup, cancels pending retries, and emits a final disconnect with `willRetry: false` (once), transitioning to **Destroyed**.
  - Repeated `destroy()` calls are safe no-ops; teardown is more robust—unsubscribe abort errors are suppressed, other failures emit warnings, and disconnect listener exceptions during destroy are handled cleanly.
- **Tests**
  - Expanded lifecycle and edge-case coverage for destruction semantics.

### DIFF
--- a/packages/sdk-ts/src/client/indexer/grpc_stream/streamV2/StreamManagerV2.ts
+++ b/packages/sdk-ts/src/client/indexer/grpc_stream/streamV2/StreamManagerV2.ts
@@ -142,6 +142,12 @@ export class StreamManagerV2<TResponse> extends EventEmitter<
   }
 
   public start(): void {
+    if (this.isDestroyed) {
+      throw new Error(
+        `Cannot start destroyed stream manager: ${this.config.id}`,
+      )
+    }
+
     if (this.state !== StreamState.Idle && this.state !== StreamState.Stopped) {
       this.emit(StreamEvent.Warn, {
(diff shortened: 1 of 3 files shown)


Frederick-88 (approved):

packages/sdk-ts/src/client/indexer/grpc_stream/streamV2/StreamManagerV2.ts
```diff
@@ -166,29 +172,72 @@ export class StreamManagerV2<TResponse> extends EventEmitter<
...
       return
     }
 
-    this.subscription.unsubscribe()
+    const subscription = this.subscription
     this.subscription = null
+
+    try {
+      subscription.unsubscribe()
+    } catch (error) {
```
ivan-angjelkoski, line 251: is there a reason why we are extracting this to a const? in looks like we can use it directly in the try block:
```
try {
  this.subscription.unsubscribe()
}
```

and then in the finally block we can set it to null

ivan-angjelkoski (approved):

PR #703, opened 2026-07-15
chore: refresh dependencies and resolve Socket findings

maxim-inj:
## Context

The workspace dependency graph included outdated direct and transitive packages
reported by Socket. Several findings came from nested dependencies whose versions
cannot be controlled through the package catalogs alone.

This PR refreshes the dependency graph, centralizes the required pnpm overrides,
and updates the SDK and wallet packages for compatibility with the newer
CosmJS, Viem, and supporting package versions.

## What changed

- Updated workspace tooling and core catalog dependencies, including Vitest,
  Lerna, Viem, Ethers, Axios, and the CosmJS packages.
- Moved pnpm overrides into `pnpm-workspace.yaml` and pinned affected transitive
  dependencies to versions that clear the Socket scan.
- Aligned the CosmJS packages on `0.39.0` and `cosmjs-types` on `0.11.0`.
- Resolved legacy cryptography dependencies through the API-compatible
  `@mapped/elliptic` package containing the upstream security fix.
- Resolved all npm `string_decoder` polyfill imports through
  `@injectivelabs/string-decoder@1.3.0` (fork: https://github.com/InjectiveLabs/string-decoder), preserving the original API and legacy
  import paths.
- Added the required Solana sysvars dependency for the Trezor dependency graph.
- Replaced the standalone `eip1193-provider` type dependency with Viem's
  EIP-1193 types and aligned wallet RPC request/client typings with the newer
  Viem definitions.
- Updated SDK compatibility for CosmJS account numbers and IBC transfer
  encoding fields.
- Made the private-key unit tests self-contained with deterministic public test
  credentials and added the wallet runtime-light source alias to Vitest.
- Regenerated `pnpm-lock.yaml` with the final dependency resolutions.

## Validation

- `socket pnpm i`
- `CI=true pnpm build` — 18 projects passed
- `pnpm type-check`
- `pnpm lint` — 18 projects passed
- `pnpm test` — 200 test files passed, 3 skipped; 1,228 tests passed, 35 skipped
- Broadcast test suite — 3 test files passed; 11 tests passed, 2 skipped
- String decoder compatibility checks against Node.js and `readable-stream`
  versions 2 and 3
- `git diff --check`

## Before / After

Before:
Screenshot_20260715_190744

After: 
Screenshot_20260715_190702

## Summary by CodeRabbit

- **New Features**
  - EVM wallet signing now consistently normalizes addresses and encodes `personal_sign` payloads across text, hex, and binary inputs.
  - Signing workflows now support larger account numbers (including `bigint`).
  - IBC transfers now preserve `memo` in Amino message payloads.
  - Browser wallet provider interfaces now offer richer event and request typings.

- **Bug Fixes**
  - Improved EVM JSON-RPC parameter formatting and address handling.
  - Account number parsing is now more direct and reliable.

- **Tests**
  - Cryptographic tests no longer rely on environment variables.

- **Chores**
  - Updated development tooling and dependency versions, plus wallet/SDK compatibility tweaks.

### DIFF
--- a/packages/sdk-ts/src/core/stargate/SigningStargateClient.ts
+++ b/packages/sdk-ts/src/core/stargate/SigningStargateClient.ts
@@ -62,7 +62,7 @@ import type {
  * @see https://github.com/cosmos/cosmos-sdk/blob/v0.42.2/x/auth/signing/sign_mode_handler.go#L23-L37
  */
 export interface SignerData {
-  readonly accountNumber: number
+  readonly accountNumber: number | bigint
   readonly sequence: number
   readonly chainId: string
 }
--- a/packages/wallets/wallet-base/src/runtime-light.ts
+++ b/packages/wallets/wallet-base/src/runtime-light.ts
@@ -155,11 +155,47 @@ export interface WalletStrategyEmitter {
   emit(event: WalletStrategyEmitterEventType, ...args: any[]): boolean
 }
 
+interface BrowserEip1193ProviderRpcError extends Error {
+  message: string
+  code: number
+  data?: unknown
+}
+
+interface BrowserEip1193ProviderMessage {
+  type: string
+  data: unknown
+}
+
+interface BrowserEip1193ProviderInfo {
+  chainId: string
+}
+
+interface BrowserEip1193RequestArguments {
+  method: string
+  params?: unknown[] | object
+}
+
 export interface BrowserEip1193Provider {
+  on(
+    event: 'connect',
+    listener: (info: BrowserEip1193ProviderInfo) => void,
+  ): void
+  on(
+    event: 'disconnect',
+    listener: (error: BrowserEip1193ProviderRpcError) => void,
+  ): void
+  on(
+    event: 'message',
+    listener: (message: BrowserEip1193ProviderMessage) => void,
+  ): void
+  on(event: 'chainChanged', listener: (chainId: string) => void): void
+  on(event: 'accountsChanged', listener: (accounts: string[]) => void): void
+  on(event: string, listener: any): void
+  once(event: string, listener: any): void
+  removeListener(event: string, listener: any): void
+  off(event: string, listener: any): void
+  request(args: BrowserEip1193RequestArguments): Promise<unknown>
   removeAllListeners(): void
-  request(args: { method: string; params?: unknown[] }): Promise<unknown>
-  on(event: string, listener: (...args: unknown[]) => void): void
-  removeListener(event: string, listener: (...args: unknown[]) => void): void
   providers?: BrowserEip1193Provider[]
   isTrust: boolean
   isKeplr: boolean
--- a/packages/wallets/wallet-base/src/runtime-light.ts
+++ b/packages/wallets/wallet-base/src/runtime-light.ts
@@ -191,9 +191,9 @@ export interface BrowserEip1193Provider {
   on(event: 'chainChanged', listener: (chainId: string) => void): void
   on(event: 'accountsChanged', listener: (accounts: string[]) => void): void
   on(event: string, listener: any): void
-  once(event: string, listener: any): void
+  once?(event: string, listener: any): void
   removeListener(event: string, listener: any): void
-  off(event: string, listener: any): void
+  off?(event: string, listener: any): void
   request(args: BrowserEip1193RequestArguments): Promise<unknown>
   removeAllListeners(): void
   providers?: BrowserEip1193Provider[]
@@ -208,8 +208,6 @@ export interface BrowserEip1193Provider {
   isOkxWallet: boolean
   isTrustWallet: boolean
 }
-
-export type BrowserEip1993Provider = BrowserEip1193Provider
 
 export interface WindowWithEip1193Provider extends Omit<
   Window,
--- a/packages/wallets/wallet-base/src/types/provider.ts
+++ b/packages/wallets/wallet-base/src/types/provider.ts
@@ -1,6 +1,43 @@
-import type { EIP1193Provider } from 'viem'
+interface BrowserEip1193ProviderRpcError extends Error {
+  message: string
+  code: number
+  data?: unknown
+}
 
-export interface BrowserEip1993Provider extends EIP1193Provider {
+interface BrowserEip1193ProviderMessage {
+  type: string
+  data: unknown
+}
+
+interface BrowserEip1193ProviderInfo {
+  chainId: string
+}
+
+interface BrowserEip1193RequestArguments {
+  method: string
+  params?: unknown[] | object
+}
+
+export interface BrowserEip1993Provider {
+  on(
+    event: 'connect',
+    listener: (info: BrowserEip1193ProviderInfo) => void,
+  ): void
+  on(
+    event: 'disconnect',
+    listener: (error: BrowserEip1193ProviderRpcError) => void,
+  ): void
+  on(
+    event: 'message',
+    listener: (message: BrowserEip1193ProviderMessage) => void,
+  ): void
+  on(event: 'chainChanged', listener: (chainId: string) => void): void
+  on(event: 'accountsChanged', listener: (accounts: string[]) => void): void
+  on(event: string, listener: any): void
+  once(event: string, listener: any): void
+  removeListener(event: string, listener: any): void
+  off(event: string, listener: any): void
+  request(args: BrowserEip1193RequestArguments): Promise<unknown>
   removeAllListeners(): void
   providers?: BrowserEip1993Provider[]
   isTrust: boolean
--- a/packages/wallets/wallet-base/src/types/provider.ts
+++ b/packages/wallets/wallet-base/src/types/provider.ts
@@ -1,4 +1,4 @@
-import type { EIP1193Provider } from 'eip1193-provider'
+import type { EIP1193Provider } from 'viem'
 
 export interface BrowserEip1993Provider extends EIP1193Provider {
   removeAllListeners(): void
--- a/packages/wallets/wallet-base/src/types/provider.ts
+++ b/packages/wallets/wallet-base/src/types/provider.ts
@@ -18,7 +18,7 @@ interface BrowserEip1193RequestArguments {
   params?: unknown[] | object
 }
 
-export interface BrowserEip1993Provider {
+export interface BrowserEip1193Provider {
   on(
     event: 'connect',
     listener: (info: BrowserEip1193ProviderInfo) => void,
@@ -34,12 +34,12 @@ export interface BrowserEip1993Provider {
   on(event: 'chainChanged', listener: (chainId: string) => void): void
   on(event: 'accountsChanged', listener: (accounts: string[]) => void): void
   on(event: string, listener: any): void
-  once(event: string, listener: any): void
+  once?(event: string, listener: any): void
   removeListener(event: string, listener: any): void
-  off(event: string, listener: any): void
+  off?(event: string, listener: any): void
   request(args: BrowserEip1193RequestArguments): Promise<unknown>
   removeAllListeners(): void
-  providers?: BrowserEip1993Provider[]
+  providers?: BrowserEip1193Provider[]
   isTrust: boolean
   isKeplr: boolean
   isRabby: boolean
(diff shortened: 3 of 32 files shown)


packages/sdk-ts/src/core/stargate/SigningStargateClient.ts
```diff
@@ -62,7 +62,7 @@ import type {
  * @see https://github.com/cosmos/cosmos-sdk/blob/v0.42.2/x/auth/signing/sign_mode_handler.go#L23-L37
  */
 export interface SignerData {
-  readonly accountNumber: number
+  readonly accountNumber: number | bigint
```
maxim-inj, line 65: This is required by the **CosmJS 0.39.0** upgrade.

  `@cosmjs/stargate` now defines `Account.accountNumber` as bigint because newer Cosmos SDK account IDs can exceed `Number.MAX_SAFE_INTEGER`.

  The old conversion:

```
  parseInt(baseAccount.accountNumber.toString(), 10)
```

  could lose precision. For example, `9007199254740993n` becomes `9007199254740992`.

  Therefore `packages/sdk-ts/src/core/accounts/AccountParser.ts:20` now preserves the protobuf value as bigint.

maxim-inj:
**Codex review:**


  - [P1] Declare cosmjs-types 0.11 in the SDK manifest — /home/xlab/dev/InjectiveLabs/injective-ts/pnpm-workspace.yaml:83-83
    When the SDK is published via pnpm pack, workspace overrides are not written into the tarball: packages/sdk-ts/package.json still contains cosmjs-types: 0.9.0 while the packed CosmJS dependencies are 0.39.0. A downstream install therefore executes the 0.9 MsgTransfer
    codec, causing MsgTransferCosmjs.toAmino() and toWeb3Gw() to omit the new encoding field that this patch's test and generated declarations guarantee; declare 0.11 directly in the SDK manifest.

  - [P2] Preserve the provider's existing request and event API — /home/xlab/dev/InjectiveLabs/injective-ts/packages/wallets/wallet-base/src/types/provider.ts:1-1
    For downstream TypeScript users of the exported BrowserEip1993Provider, viem's type drops once/off, restricts event names, and limits request to its EIP-1474 schema. Existing code calling wallet-specific methods or those event helpers compiled against the old public
    contract and now fails type checking even though injected wallet objects still expose them; preserve compatible request and event overloads while removing the old dependency.

  - [P2] Raise linkify-it to the DoS-fixed release — /home/xlab/dev/InjectiveLabs/injective-ts/pnpm-workspace.yaml:55-55
    When Markdown-It linkification is enabled for untrusted Markdown, this override forces 5.0.1 even though Markdown-It 14.3.0 requires ^5.0.2. In 5.0.1, repeated mailto: strings and long URL-auth prefixes trigger quadratic rescanning, allowing crafted input to tie up the
    process; use 5.0.2 or newer within the required major.

maxim-inj:
One more issue

```
  - [P2] Omit unsupported encoding from Web3GW payloads — /home/xlab/dev/InjectiveLabs/injective-ts/packages/sdk-ts/src/core/modules/ibc/msgs/MsgTransferCosmjs.spec.ts:82-82
    When MsgTransferCosmjs.toWeb3Gw() is sent to IndexerGrpcTransactionApi.prepareTxRequest, the current Injective ibc.applications.transfer.v1.MsgTransfer schema does not define this field, so strict JSON unmarshalling fails with unknown field "encoding". This assertion
    codifies a runtime-breaking default introduced by cosmjs-types@0.11; strip encoding from Amino/Web3GW objects instead of spreading the codec default.
```

This is something to address in our `ibc.applications.transfer.v1.MsgTransfer` schema but I'll fix for backwards-compat for now.

pnpm-workspace.yaml
```diff
@@ -3,22 +3,100 @@ packages:
...
+overrides:
+  zod: 3.24.2
+  rxjs: 7.8.2
+  rpc-websockets: 9.3.5
+  next: file:./etc/noop
+  '@ethereumjs/tx': ^4.1.1
+  '@solana/web3.js': 1.98.1
+  '@solana-program/token-2022@0.4.2>@solana/sysvars': 2.3.0
+  '@ethereumjs/common': 3.2.0
+  '@coinbase/cdp-sdk': 1.46.0
```
ivan-angjelkoski: ⚠️ **Potential issue — CDP 1.46.0 omits a runtime dependency**

`@coinbase/cdp-sdk@1.46.0` imports `bs58` from the eagerly loaded `CdpClient` graph but does not declare it. `@base-org/account`'s Node entrypoint exports payment helpers that import `CdpClient`, so consumers using pnpm's default isolated layout can fail during module instantiation with `ERR_MODULE_NOT_FOUND`. This workspace's shameful hoisting masks the omission because another Solana subtree happens to install `bs58`; CDP 1.48.3 declares the dependency.

<details>

```suggestion
  '@coinbase/cdp-sdk': 1.48.3
```

</details>

<details>

```

In `@pnpm-workspace.yaml` around line 15, the 1.46.0 override selects a
CDP package that imports `bs58` but does not declare it. This can make
`@base-org/account`'s Node payment entrypoint fail under pnpm's default
isolated layout even though shameful hoisting masks the omission here.
Upgrade to 1.48.3 or a newer verified release that declares `bs58`, and
ensure its `zod` dependency is not forced below its declared range.
Regenerate `pnpm-lock.yaml` and verify a clean Node import without hoisting.
```

</details>

pnpm-workspace.yaml
```diff
@@ -3,22 +3,100 @@ packages:
...
+  rollup: 4.62.2
+  sigstore: 4.1.1
+  # Injective fork avoids Socket's underscore parsing issue while preserving the original API.
+  string_decoder: npm:@injectivelabs/string-decoder@1.3.0
+  tar@6: 7.5.16
+  tar@7: 7.5.16
+  tmp: 0.2.7
+  ua-parser-js: 2.0.10
+  uuid@8: 11.1.1
+  uuid: 11.1.1
```
ivan-angjelkoski, line 78: 🛠️ **Refactor suggestion — UUID 11 breaks `http-auth` Digest challenges**

The blanket override resolves `http-auth@3.1.3`'s `uuid@^3.0.0` dependency to 11.1.1. That package executes `const uuid = require('uuid')` and later calls `uuid()`, but UUID 11 returns an exports object, so the first Digest challenge throws `TypeError: uuid is not a function`. `live-server` currently exercises Basic auth only, which lets this incompatible resolution pass validation; upgrade `http-auth` to 4.2.1 or otherwise keep its UUID API compatible before forcing UUID 11.

<details>

```

In `@pnpm-workspace.yaml` around line 77, the blanket UUID override forces
`http-auth@3.1.3` from UUID 3 to UUID 11. That version of `http-auth` calls
the result of `require("uuid")` as a function, while UUID 11 returns an
exports object, causing Digest challenge generation to throw. Upgrade or
patch `http-auth` to use `v4`, or scope the override without restoring the
insecure UUID release. Regenerate the lockfile and smoke-test Basic and
Digest challenges.
```

</details>

packages/wallets/wallet-base/src/runtime-light.ts
```diff
@@ -155,11 +155,47 @@ export interface WalletStrategyEmitter {
...
+  on(
+    event: 'message',
+    listener: (message: BrowserEip1193ProviderMessage) => void,
+  ): void
+  on(event: 'chainChanged', listener: (chainId: string) => void): void
+  on(event: 'accountsChanged', listener: (accounts: string[]) => void): void
+  on(event: string, listener: any): void
+  once(event: string, listener: any): void
+  removeListener(event: string, listener: any): void
+  off(event: string, listener: any): void
```
ivan-angjelkoski: 🧹 **Nitpick (non-blocking) — Keep new light-provider methods optional**

The public `runtime-light` interface previously accepted providers implementing `request`, `on`, `removeListener`, and `removeAllListeners`. Making `once` and `off` mandatory now rejects existing custom wrappers even though neither method is used through this interface. Keeping the newly added methods optional preserves structural compatibility while still describing providers that expose them.

<details>

```suggestion
  once?(event: string, listener: any): void
  removeListener(event: string, listener: any): void
  off?(event: string, listener: any): void
```

</details>

<details>

```

In `@packages/wallets/wallet-base/src/runtime-light.ts` around lines
194 - 196, newly required `once` and `off` methods break custom providers
that satisfied the previous public light-provider shape. Make those two
new methods optional while keeping `removeListener` required. Add a
compile-time fixture proving the previous provider shape remains assignable
and confirm wallet event registration still type-checks.
```

</details>

ivan-angjelkoski (commented):
## Review summary

Three findings: one critical, one medium, and one minor.

- **Critical** — The CDP 1.46.0 override selects a package that imports undeclared `bs58`, which can break isolated pnpm consumers.
- **Medium** — The blanket UUID 11 override breaks `http-auth@3.1.3` Digest challenge generation.
- **Minor** — Newly mandatory `once` and `off` methods narrow the public `runtime-light` provider type.

Security: no credential leakage or signing-validation bypass was found. Performance: no actionable findings.

Positives: the CosmJS bigint migration preserves account-number precision, the EVM signing changes now validate addresses and encode payloads consistently, and the dependency refresh is backed by broad build, lint, and test validation.

packages/wallets/wallet-base/src/types/provider.ts
```diff
@@ -1,8 +1,45 @@
...
+  chainId: string
+}
+
+interface BrowserEip1193RequestArguments {
+  method: string
+  params?: unknown[] | object
+}
 
-export interface BrowserEip1993Provider extends EIP1193Provider {
+export interface BrowserEip1193Provider {
```
ivan-angjelkoski, line 21: 🛠️ **Refactor suggestion — Removing the exported `BrowserEip1993Provider` breaks downstream consumers**

`BrowserEip1993Provider` was previously exported from this module and re-exported through the package's main entry (`src/index.ts` → `types/index.ts` → `provider.ts`), so it is part of `@injectivelabs/wallet-base`'s public type surface. Renaming it to `BrowserEip1193Provider` and dropping the alias that existed in `runtime-light.ts` is a breaking change: any external consumer importing the old name will fail to compile after upgrading. The spelling fix itself is correct (EIP-**1193**), but for a published package a hard rename without a deprecation shim forces a coordinated major bump across every downstream repo. A `@deprecated` type alias preserves compatibility for one release cycle at zero runtime cost (types are erased at build).

<details>

```suggestion
/** @deprecated Use `BrowserEip1193Provider`. The `1993` spelling was a typo. */
export type BrowserEip1993Provider = BrowserEip1193Provider

export interface BrowserEip1193Provider {
```

</details>

<details>

```

In `@packages/wallets/wallet-base/src/types/provider.ts` around line 21,
the exported type `BrowserEip1993Provider` was renamed to
`BrowserEip1193Provider` and the old name was removed entirely.
`BrowserEip1993Provider` was re-exported from the package's main entry
(`src/index.ts` -> `types/index.ts` -> `provider.ts`), so external
consumers importing the old name will break on upgrade. Re-add a
`@deprecated` type alias (`export type BrowserEip1993Provider =
BrowserEip1193Provider`) so downstream code keeps compiling for one
release cycle; the rename itself is correct. Also check whether other
published InjectiveLabs packages import the old name and update them.
```

</details>

packages/wallets/wallet-base/src/runtime-light.ts
```diff
@@ -155,11 +155,47 @@ export interface WalletStrategyEmitter {
...
+
+interface BrowserEip1193ProviderInfo {
+  chainId: string
+}
+
+interface BrowserEip1193RequestArguments {
+  method: string
+  params?: unknown[] | object
+}
+
```
ivan-angjelkoski, line 177: 🧹 **Nitpick (non-blocking) — Provider interfaces are duplicated verbatim from `types/provider.ts`**

The four helper interfaces (`BrowserEip1193ProviderRpcError`, `BrowserEip1193ProviderMessage`, `BrowserEip1193ProviderInfo`, `BrowserEip1193RequestArguments`) plus `BrowserEip1193Provider` are now declared identically in both `runtime-light.ts` and `types/provider.ts`. They must be kept in sync by hand — adding a new event overload or wallet flag in one file but not the other will silently diverge the two entry points. Since these are type-only declarations (erased at build time), `runtime-light.ts` could import them from `./types/provider.js` without growing the runtime bundle, once those helpers are exported from `provider.ts`. I understand `runtime-light.ts` is intentionally self-contained, so this is at the author's discretion.

<details>

```

In `@packages/wallets/wallet-base/src/runtime-light.ts` around lines
158 - 177, the helper interfaces `BrowserEip1193ProviderRpcError`,
`BrowserEip1193ProviderMessage`, `BrowserEip1193ProviderInfo`, and
`BrowserEip1193RequestArguments`, plus `BrowserEip1193Provider`, are
duplicated verbatim from `types/provider.ts`. The two copies must be
kept in sync manually or the `/light` and main entry points will
silently diverge. Consider exporting the shared helper interfaces from
`types/provider.ts` and importing them type-only into
`runtime-light.ts`; type-only imports are erased at build time so they
do not affect the light bundle size. Only do this if it does not break
the intended isolation of the light entry point.
```

</details>

ivan-angjelkoski (commented):
## Review summary

Two findings, none blocking:

- **Medium** — `types/provider.ts`: removing the exported `BrowserEip1993Provider` (renamed to `BrowserEip1193Provider`) is a breaking change to `@injectivelabs/wallet-base`'s public type surface; a `@deprecated` alias eases downstream migration.
- **Minor** — `runtime-light.ts`: the provider helper interfaces are now duplicated verbatim from `types/provider.ts`, creating a manual sync burden between the two entry points.

**Positives:** This is a clean, thorough dependency refresh. The EVM signing-payload changes are real improvements — `personal_sign` now correctly emits hex for text/hex/bytes inputs, addresses are checksummed via `getAddress`, and the new `strategy.spec.ts` covers all three input shapes deterministically. The CosmJS 0.39 alignment is done correctly (`accountParser` now returns a `bigint` `accountNumber` to match `@cosmjs/stargate`'s `Account`, and `SignerData` is widened to `number | bigint`). The `BrowserEip1993`→`BrowserEip1193` typo fix is applied consistently with zero dangling references, the IBC amino encoding switched to explicit field listing to avoid leaking new proto fields, and the Socket-driven overrides are well-commented (elliptic fork, string-decoder fork). Security: no findings (the test mnemonic/private key are canonical public BIP39 test vectors, not secrets). Performance: no findings.

ThomasRalee (approved):

